Issue #2, opened 2019-02-26
Run automated / CI testing


mphoward:
Now that azplugins is available publicly on github, I need to find a new testing / CI platform since the old Jenkins server is no longer available. CircleCI seems like a good option at least for the CPU tests. (There aren't any GPUs available for execution.) To get those tests up and running, I need:

- [x] Build docker containers based on [HOOMD's ci images](https://hub.docker.com/r/glotzerlab/ci/).

These containers will have a few different build configurations of hoomd installed, and then the plugins can be built externally to those to save on compile time. We probably need two versions of hoomd: (1) the minimum supported version and (2) "bleeding edge" hoomd (either the latest tag or maint) to look out for bugs. (2) might be better suited to a download-and-build since this code may continually change.

- [x] Setup a CircleCI config file to build and run the various test configurations on the CPU.
- [ ] Add testing for the documentation once it has been built (#1)

A long-term testing option is needed for the GPU if significant code development is going to continue on this project.

mphoward:
Somewhere, we need to test a mix of CUDA versions, as some compilation errors only show up in CUDA 9 (see #4).

mphoward:
Will defer doc testing until #1 is completed/

mphoward:
Implemented on master and mostly polished off in #7 .

PR #7, opened 2019-04-02
Fix import statements in azplugins modules

### LINKED ISSUE #6: Fix python3 compatibility
The import style used is only compatible with python2 and raises ImportError exceptions in python3. Replace statements like
```
import analyze
```
with
```
from azplugins import analyze
```
This is another example where #2 could have caught this issue.

(Reported by Arash)

mphoward:
Arash has tested this branch and confirms that it works for him. I've also run the unit tests for a selected python3 installation.

Resolves #6

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,7 +23,29 @@ commands:
           command: |
             mkdir build
             cd build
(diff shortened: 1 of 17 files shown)


astatt:
I've reproduced the bug with python 3.6.7 (anaconda3)

mphoward:
Also resolves #7 and resolves #2 (other than doc testing, tbd).

Issue #9, opened 2019-06-13
Support HOOMD 2.6.0


mphoward:
HOOMD 2.6.0 introduces confined MPCD streaming geometries, which leads to symbol collisions with `BoundaryCondition.h` and `SlitGeometry.h`. (These objects were copied from that implementation while it was being developed, and actually reside in the `mpcd` namespace.)

There are three options to fix this:
1. Easiest: bump the minimum version of HOOMD for azplugins to be 2.6.0 and always require an MPCD build. Then, remove `BoundaryConditions.h` and `SlitGeometry.h` from azplugins and use the mpcd headers from HOOMD.
2. Harder: change the namespace in `BoundaryConditions.h` and `SlitGeometry.h` to be azplugins rather than mpcd. This should avoid the naming collision in the binary, but you might need to also change the python export name (prepend "AZP" or something).
3. Hardest: keep the files as is, and conditionally include / compile them depending on the version of HOOMD that azplugins has been built against. At the python level, this means conditionally importing as well.

Option 1 is probably the most straightforward approach for now since azplugins does not promise backward compatibility with any given version of HOOMD (the minimum is stated in the release notes).

The reason to require an MPCD build now is that the `SlitGeometry` is used in the `BounceBackNVE` integrator, which is intended to be run with things like DPD, regular LJ particles, etc., and without the mpcd component of HOOMD, these will be missing. We should add an error like this to CMake for `BUILD_MPCD`:

https://github.com/mphoward/azplugins/blob/582513a0e36d8e5541290312268bdf4a1b524335/azplugins/CMakeLists.txt#L15-L17

That way, the conditional compilation of the MPCD component could possibly be enabled again at a later time.

- [x] Build a new CI container testing HOOMD 2.6.0 in order to confirm these changes work.
- [x] Update CMake to force MPCD build.
- [x] Strip out azplugins versions of these MPCD files, and use the native HOOMD 2.6.0 ones.

mphoward:
The new test container has been built. As expected, builds fail for the develop branch when the python module is imported due to naming clashes. So, I merged it into this one, and tests are running now. `py3-internal-nompcd` should be the only failure for this build, and I will disable it after we confirm this working.

mphoward:
Resolved by #12

PR #12, opened 2019-06-19
Fix/hoomd2.6

astatt:
This fix removes the name clash described in Issue #9 Support HOOMD 2.6.0. Minimum HOOMD version is bumped to HOOMD 2.6.0 as well.


azplugins/BounceBackNVEGPU.cu
```diff
@@ -10,7 +10,10 @@
  */
 
 #include "BounceBackNVEGPU.cuh"
-#include "BounceBackGeometry.h"
```
mphoward: I'm in favor of leaving the mpcd headers in `BounceBackGeometry.h` and not explicitly including them individually here, so you can revert this change. Otherwise, remove the headers from `BounceBackGeometry.h`.

azplugins/CMakeLists.txt
```diff
@@ -15,6 +15,10 @@ endif(NOT DEFINED HOOMD_EXTERNAL_BUILD)
 if (NOT BUILD_MD)
     message(FATAL_ERROR "azplugins requires hoomd installation with MD component.")
 endif (NOT BUILD_MD)
+# the md component is always required
```
mphoward: `md` -> `mpcd`

azplugins/module.cc
```diff
@@ -55,6 +55,10 @@ namespace py = pybind11;
 
 /* Integrators */
 #include "BounceBackGeometry.h"
+#ifdef ENABLE_MPCD
```
mphoward: Again, if you stick these in `BounceBackGeometry.h`, no need to include here.

azplugins/BounceBackNVEGPU.cu
astatt: I've put the headers in `BounceBackGeometry.h` and removed them from  `BounceBackNVEGPU.cu` and module.cc .

Issue #10, opened 2019-06-13
Guarantee backward compatibility with HOOMD


mphoward:
Right now, azplugins does not guarantee any backwards compatibility with HOOMD, which can be somewhat annoying when you need to recompile for a small feature change.

HOOMD 3.0 is going to break current plugin builds anyway as it will improve the underlying CMake functionality, so I think this is where we would start promising backward compatibility. This will be easier to implement at that point since HOOMD 3.0 will supply the CMake files needed to deduce its version, and so you can do things like "include this if version >= else...".

This will be a feature of the azplugins 1.0 major release.

mphoward:
There is no backwards compatibility issue with HOOMD 4, but we can reconsider this if there are API breaking changes moving forward.

PR #15, opened 2019-08-20
Slit pore bounce back integrator

mphoward:
This PR adds support for the NVE bounce back integrator in the slit pore geometry that will be added to HOOMD 2.7 (https://github.com/glotzerlab/hoomd-blue/issues/440). As such, it shouldn't be merged until that is released, and we will also need to start testing that version of hoomd in CI.

To try to start guaranteeing compatibility with different HOOMD APIs (#10), this PR also implements an experimental decorator interface. The idea is that on the python side, we will decorate classes with things like
```
@api.require('2.7.0')
class slit_pore(...):
```
which will do an automatic check against the tagged version of HOOMD, and raise an error if the minimum is not met. On the C++ side, we can do conditional compilation. There are preprocessor defines for `HOOMD_VERSION_MAJOR` and `HOOMD_VERSION_MINOR`. I propose letting the azplugins headers define their own API keys that mimic the python command naming as needed:
```
#if (HOOMD_VERSION_MAJOR >= 2) && (HOOMD_VERSION_MINOR >= 7)
#define AZPLUGINS_API_INTEGRATE_SLIT_PORE
#endif
```
For now, I did this in the header, but we could also create some sort of `api.h` that lets us define all of these in one place. The only potential issue with this is when `*.cc` and `*.h` files should be completely skipped (and not installed), which is a CMake issue.

The decorator api needs testing and I might change some things as I try using this in production, but opening this now anyway.

### DIFF
--- a/azplugins/BounceBackGeometry.h
+++ b/azplugins/BounceBackGeometry.h
@@ -18,6 +18,11 @@
 #include "hoomd/mpcd/BoundaryCondition.h"
 #include "hoomd/mpcd/SlitGeometry.h"
 
+#if (HOOMD_VERSION_MAJOR >= 2) && (HOOMD_MINOR_VERSION >= 7)
+#define AZPLUGINS_API_INTEGRATE_SLIT_PORE
+#include "hoomd/mpcd/SlitPoreGeometry.h"
+#endif
+
 #ifndef NVCC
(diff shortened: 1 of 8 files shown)


mphoward:
This integrator has been implemented in HOOMD, so this PR is no longer needed. However, I want to save the API decorator for future, so I will pull that code out into a separate branch before closing and deleting.

PR #17, opened 2019-10-15
Implicit droplet evaporator

mphoward:
This PR implements the moving harmonic potential that mimics evaporation in a spherical geometry, as in [Liu et al.](https://doi.org/10.1021/acsnano.9b00459) Since the original command for the thin film evaporator was `evaporate.implicit` and the class structures are highly similar, I decided to add a `geometry` option to this command that can be either `'film'` (default, original behavior) or `'droplet'` (new geometry). This supports extension to other geometries like cylinders. On the C++ side, there is now a base `ImplicitEvaporator`, from which the geometries and GPU versions derive. Technically, this is probably better achieved using templates to compute the distance and normal to the interface rather than subclassing, but this was faster to implement. The code can be refactored later if we want without changing the public API.

### DIFF
--- a/azplugins/ImplicitEvaporator.cc
+++ b/azplugins/ImplicitEvaporator.cc
@@ -20,8 +20,6 @@ ImplicitEvaporator::ImplicitEvaporator(std::shared_ptr<SystemDefinition> sysdef,
                                        std::shared_ptr<Variant> interf)
         : ForceCompute(sysdef), m_interf(interf), m_has_warned(false)
     {
-    m_exec_conf->msg->notice(5) << "Constructing ImplicitEvaporator" << std::endl;
-
     // allocate memory per type for parameters
     GPUArray<Scalar4> params(m_pdata->getNTypes(), m_exec_conf);
     m_params.swap(params);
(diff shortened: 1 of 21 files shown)


azplugins/ImplicitEvaporator.cc
```diff
@@ -32,67 +30,23 @@ ImplicitEvaporator::ImplicitEvaporator(std::shared_ptr<SystemDefinition> sysdef,
...
+ * Deriving classes should implement the actual force compute method.
  */
 void ImplicitEvaporator::computeForces(unsigned int timestep)
     {
-    PDataFlags flags = this->m_pdata->getFlags();
+    PDataFlags flags = m_pdata->getFlags();
     if (!m_has_warned && (flags[pdata_flag::pressure_tensor] || flags[pdata_flag::isotropic_virial]))
         {
-        m_exec_conf->msg->warning() << "ImplicitEvaporator does not compute its virial contribution, pressure may be inaccurate" << std::endl;
+        m_exec_conf->msg->warning() << "ImplicitPlaneEvaporator does not compute its virial contribution, pressure may be inaccurate" << std::endl;
```
mphoward: Revert this, as this is in the base clase.

azplugins/evaporate.py
```diff
@@ -42,6 +43,15 @@ class implicit(hoomd.md.force._force):
     (the force is continuous) and also that :math:`F_g` scales with the cube of the
     particle radius.
 
+    The drying *geometry* determines the physical meaning of *z* and *H*. Currently,
+    it can be:
+
+    - *film*: A drying thin film (planar geometry), where *z* is the particle's
+              z-coordinate of the particle and *H* is then the film height.
```
mphoward:
```suggestion
              z-coordinate and *H* is then the film height.
```

PR #20, opened 2019-11-19
Shrinking sphere variant

mphoward:
The shrinking sphere is defined as in [this paper](http://doi.org/10.1021/acsnano.9b00459) so that it has a constant rate of surface area reduction. This is a useful variant for drying in a spherical droplet geometry, but might be useful elsewhere. The variant code is pretty lightweight anyway.

### DIFF
--- a/azplugins/VariantSphereArea.cc
+++ b/azplugins/VariantSphereArea.cc
@@ -4,35 +4,29 @@
 // Maintainer: mphoward
 
 /*!
- * \file VariantSphere.cc
+ * \file VariantSphereArea.cc
  * \brief Defines a variant for a shrinking / growing sphere radius.
  */
 
-#include "VariantSphere.h"
+#include "VariantSphereArea.h"
 
 namespace azplugins
 {
 /*!
  * \param R0 Initial radius.
- * \param alpha Rate of surface reduction (units are per timestep).
- *
- * The parameters are converted into quantities that are more useful internally.
- * The units of alpha are implicitly per-timestep.
+ * \param alpha Rate of surface reduction (units are area per timestep).
  */
-VariantSphere::VariantSphere(double R0, double alpha)
+VariantSphereArea::VariantSphereArea(double R0, double alpha)
     {
-    // precompute quantities for easy evaluation later
     m_R0_sq = R0*R0;
-
-    const Scalar factor = 0.3848347315591266;   // (6/pi)^(2/3) / 4
-    m_k = factor*alpha;
+    m_k = alpha/(4.*M_PI);
     }
 
 /*!
  * \param timestep Current simulation timestep.
  * \returns Current radius of sphere.
  */
-double VariantSphere::getValue(unsigned int timestep)
+double VariantSphereArea::getValue(unsigned int timestep)
     {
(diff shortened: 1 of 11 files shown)


azplugins/VariantSphereArea.cc
```diff
@@ -0,0 +1,57 @@
...
+ * \brief Defines a variant for a shrinking / growing sphere radius.
+ */
+
+#include "VariantSphereArea.h"
+
+namespace azplugins
+{
+/*!
+ * \param R0 Initial radius.
+ * \param alpha Rate of surface reduction (units are area per timestep).
```
ara137: Should we use "area per timestep" or "area per time" as the unit for alpha? I would have used the latter (if possible), since then alpha does not need to be tuned if the timestep changes.

azplugins/VariantSphereArea.h
```diff
@@ -0,0 +1,63 @@
...
+ * \f[
+ * R(t) = \sqrt{R(0)^2 - (\alpha/4\pi) t}
+ * \f]
+ *
+ * where \f$\alpha\f$ is the rate of surface area reduction per time.
+ *
+ * To be physical, \f$R(t)\f$ is never negative; it is fixed to 0 if \f$t\f$ would make the
+ * square-root argument negative.
+ *
+ * Setting \f$\alpha < 0\f$ will cause the sphere to expand.
```
ara137: Do we need to worry about potential overflows if the radius expands to infinity?

azplugins/test-py/test_variant_sphere_area.py
```diff
@@ -0,0 +1,50 @@
...
+    def test_create(self):
+        """Test creation of sphere objects."""
+        # floats
+        s = azplugins.variant.sphere_area(R0=10., alpha=0.5)
+        self.assertAlmostEqual(s.R0, 10.)
+        self.assertAlmostEqual(s.alpha, 0.5)
+        self.assertTrue(s.cpp_variant is not None)
+
+        # ints
+        azplugins.variant.sphere_area(R0=10, alpha=1)
```
ara137: Is the test incomplete? Do we not want to test whether the values have been correctly set?

azplugins/VariantSphereArea.cc
mphoward: No, this is not possible with the current design of HOOMD. Variants are only aware of the current integer *timestep*, as the (floating-point) size of a timestep is associated to an integrator.

azplugins/VariantSphereArea.h
mphoward: The value returned is a double, so it will not overflow. A variant does not actually imply any particular physical quantity (although this one does), so it is also OK for the variant to, say, expand beyond the size of the box. It is the user's responsibility to make sure this value is sensible.

azplugins/test-py/test_variant_sphere_area.py
mphoward: Good to check, but no, I only added this test to ensure that the implicit typecast from int to float worked correctly in the constructor (it would raise an exception otherwise). Sometimes this fails with pybind11.

PR #24, opened 2019-12-19
Generalized wall restraints

mphoward:
This PR generalizes `restrain.plane` to also support other native `WallData` geometries like cylinders and spheres. This is achieved by templating the class on the wall object and using the built-in methods to compute normal vectors. The result is two new methods `restrain.cylinder` and `restrain.sphere`.


azplugins/WallRestraintComputeGPU.h
```diff
@@ -0,0 +1,104 @@
...
+
+#include "WallRestraintCompute.h"
+#include "WallRestraintComputeGPU.cuh"
+
+#include "hoomd/Autotuner.h"
+#include <hoomd/extern/pybind/include/pybind11/pybind11.h>
+
+namespace azplugins
+{
+//! Applies a harmonic force relative to a WallDAta object for a group of particles on the GPU
```
astatt: WallDAta --> WallData

mphoward:
Thanks! The current design of the `restrain` module does not restrict how many restraints can be applied to the same particle. This is for two reasons: (1) there is no central structure registering which groups are attached to which restraints (as in the `Integrator`) so this is not technically possible with the current design, and (2) there are some types of restraints that can be intended to operate on the same group (e.g., position + orientation for doing TI).

As `ForceCompute`s, there is no technical reason that multiple restraints cannot safely operate on the same group (since the forces are additive), but the physics could become nonsensical. Or, they could be interesting, for example, combining a cylinder with a plane could confine a particle to a ring.

afaik, the wall potentials are also completely additive? `md.wall.group` is converted to a struct that packs a bunch of walls together, and then the potential evaluator loops over them. The main difference is that the walls act on all particles in the system by type, rather than on a group.

Basically, is any of this helpful to put in the documentation, or is it more confusing than clarifying?

astatt:
Let's add something to the python documentation along the lines: "Restrains/Force computes are additive and multiple restrains and force computes can be used in the same simulation. The user is responsible for ensuring that the system is defined and initialized properly."

mphoward:
Docs updated. I've also used this branch in production simulations on TACC GPUs, so once the unit tests clear again (and I make sure the production output isn't nonsense) this will be ready to merge.

PR #26, opened 2020-02-19
Feature/velocity profile

astatt:
Simple helper class to determine 1D velocity profiles, e.g. v_z(x) during the simulation. Future improvements could/should include:

1. check for box size fluctuations for pure md simulations (mpcd systems error out anyway), parameter checks (will try to do that soon)
2. check if the path for the filename exists/ is writable 
3. allow measuring multiple velocity components along the same spatial direction with one instance of the class, e.g. measure v_x(x), v_y(x),.. 
4. measure temperature profiles
5. 2D profiles (probably separate class)


mphoward:
Do you want me to have a look at this now, or are you still working on some parts of it?

mphoward:
I dug into some of my files, and I made some changes to the API. I think this signature simplifies some of the usage. In terms of the I/O, I think it should be left up to the user to decide how to do the averaging, since this is a little complicated. This object just does the running average, and then it could be wrapped by another type of analyzer that decides if writing should be done periodically or only at the end.

Should we add a `reset` method that zeros the internal counters?

This also needs testing--I'm sure there are typos in it. :-) Let me know what you think!

astatt:
I'll do some user-level testing and carefully reading on the weekend, will not have time before that. The API is much simpler than what I had, which is probably good. A `reset` call would be nice, gives the user more flexibility in how they want to use it. 

If we ever want to implement a 1D temperature (for evaporation) or a 2D velocity (for channel flows) version of this, I think your API is more flexible as well.

azplugins/flow.py
```diff
@@ -653,3 +654,87 @@ def set_params(self, group=None, Nswap=None, width=None, target_momentum=None):
...
+
+        # edges span the range
+        edges = np.linspace(range[0], range[1], bins+1)
+        self.centers = 0.5*(edges[:-1]+edges[1:])
+        self._dx = edges[1:]-edges[:-1]
+        self.area = area
+
+        # profiles are initially empty
+        self._velocity = np.zeros(self.bins)
+        self._counts = np.zeros(self.bins)
```
mphoward: Should this be `dtype=np.int64` to do counting in integer math?

azplugins/flow.py
```diff
@@ -653,3 +654,87 @@ def set_params(self, group=None, Nswap=None, width=None, target_momentum=None):
...
+
+        """
+        hoomd.util.quiet_status()
+        snap = self.system.take_snapshot()
+        hoomd.util.unquiet_status()
+
+        x = snap.particles.position[:,self.bin_axis]
+        v = snap.particles.velocity[:,self.flow_axis]
+
+        _counts,_ = np.histogram(x, bins=self.num_bins, range=self.range)
```
mphoward: Typo: `self.num_bins` -> `self.bins`

azplugins/flow.py
```diff
@@ -653,3 +654,87 @@ def set_params(self, group=None, Nswap=None, width=None, target_momentum=None):
...
+        self._counts += _counts
+
+        _velocity,_ = np.histogram(x, bins=self.num_bins, range=self.range, weights=v)
```
mphoward: Typo: `self.num_bins` -> `self.bins`

azplugins/flow.py
```diff
@@ -653,3 +654,87 @@ def set_params(self, group=None, Nswap=None, width=None, target_momentum=None):
...
+        self._velocity += _velocity
+
+        self.samples += 1
+
+    @property
+    def density(self):
+        r"""The current average density profile."""
+        if self.samples > 0:
+            return np.divide(self._counts, self._dx*self.area)/self.samples
```
mphoward: Don't need to use `np.divide` here, as `where` is not used.

mphoward:
OK, thanks! I caught some typos and also added the reset method. Let me know when you try out the code, and feel free to add comments to the PR if anything needs to be fixed. If it's successful, we can write a small unit test to make sure it's binning and averaging correctly, and then this should be good to go.

azplugins/flow.py
mphoward: Probably not necessary.

mphoward:
I had forgotten about this PR :-) Do you have any additional feedback on it, or is it ready to go?

astatt:
Do you think this should have a unit test?

mphoward:
We should probably test it, but we can do it a lazy way where we initialize a system with a couple particles in it with fixed velocities and then make sure we get the right values back.

astatt:
I've added a very simple lazy unit test, where I just computed the expected histograms by hand and made sure it records the right values. If you think that's good enough for a unit test, this can be merged.

mphoward:
Thank you! Yes, this is exactly the level of test I had in mind. It looks like there are some MPI errors, which I think are in both the base code and the unit tests. I will correct those, and then this should be good to merge.

PR #32, opened 2020-07-03
Hertz pair potential

arjungoswami:
Implements the Hertz pair potential into azplugins. The Hertz potential, given as $\epsilon (1 - r / \sigma)^{5/2}$ for $r < \sigma$ and $0$ for $r> \sigma$ is a purely repulsive potential, bounded at $r = 0$, monotonically decreasing as a power-law form, and vanishing at the cutoff $\sigma$. It is a physically reasonable model for soft molecules and thus its addition into azplugins is motivated.


azplugins/test-py/test_pair_hertz.py
```diff
@@ -0,0 +1,173 @@
...
+        hertz.pair_coeff.set('A', 'A', epsilon=1.0, r_cut=2.5, r_on=2.0)
+        hertz.update_coeffs()
+
+    # test missing coefficients
+    def test_set_missing_epsilon(self):
+        hertz = azplugins.pair.hertz(r_cut=3.0, nlist = self.nl)
+        hertz.pair_coeff.set('A', 'A')
+        self.assertRaises(RuntimeError, hertz.update_coeffs)
+
+    # test missing coefficients
```
mphoward, line 36: `coefficients` -> `pair` in this comment

azplugins/test-py/test_pair_hertz.py
```diff
@@ -0,0 +1,173 @@
+# Copyright (c) 2018-2020, Michael P. Howard
+# This file is part of the azplugins project, released under the Modified BSD License.
+
+# Maintainer: arjunsg2
+
+from hoomd import *
```
mphoward, line 6: This may be copied from elsewhere, but please change this to:
```py
import hoomd
```
so that you aren't importing everything from the hoomd namespace. You may need to modify some of the things below (`data.`, `init.`, `context.`).

azplugins/test-py/test_pair_hertz.py
```diff
@@ -0,0 +1,173 @@
...
+        self.assertAlmostEqual(f0[0],-F,3)
+        self.assertAlmostEqual(f0[1],0)
+        self.assertAlmostEqual(f0[2],0)
+
+        self.assertAlmostEqual(f1[0],F,3)
+        self.assertAlmostEqual(f1[1],0)
+        self.assertAlmostEqual(f1[2],0)
+
+        hertz = azplugins.pair.hertz(r_cut=2.05, nlist = self.nl)
+        hertz.pair_coeff.set('A','A', epsilon=2.0)
```
mphoward, line 134: I would recommend testing with a different value of epsilon here to make sure it is set correctly.

azplugins/PairEvaluatorHertz.h
```diff
@@ -0,0 +1,103 @@
...
+         *
+         * The calculation does not occur if the pair distance is greater than the cutoff
+         * or if the potential is scaled to zero.
+         */
+        DEVICE bool evalForceAndEnergy(Scalar& force_divr, Scalar& pair_eng, bool energy_shift)
+            {
+            if (rsq < rcutsq && epsilon != 0)
+                {
+                Scalar r = sqrt(rsq);
+                Scalar rcut = sqrt(rcutsq);
```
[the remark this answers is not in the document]
arjungoswami: Sorry am not entirely sure what you mean here. Should I simply replace all instances of 'sqrt()' with 'fast::sqrt()'? I know :: is the scope resolution operator but am not sure how the 'fast' namespace in hoomd/HOOMDMath.h would be shared in the PairEvaluator.h file. Currently I have in line 15 a '#include <math.h>' statement that I think enables me to use the ordinary sqrt() as it is now.

Please be patient with me as I am fairly new to using hoomd, C++ programming, and GitHub in general :)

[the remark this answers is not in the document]
mphoward: No problem! There is no need to include `<math.h>`, and `hoomd/HOOMDMath.h` is already included for you somewhere way above this header. So, you can just replace `sqrt()` by `fast::sqrt()`, and it should work.

azplugins/test-py/test_pair_hertz.py
arjungoswami: I tried to do it this way, but after I did this and changed the functions below to include the extra 'hoomd.' the unit test would not pass. Unfortunately I could not isolate the source of this discrepancy (even after trying to add the 'hoomd.' to other things you didn't mention such as hoomd.run(), hoomd.group(), etc.). As you stated I copied this from elsewhere and was using this as a template.

azplugins/PairEvaluatorHertz.h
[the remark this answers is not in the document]
arjungoswami: Thank you! I have been able to successfully implement all your suggested revisions (with just one exception- the import hoomd statement in the unit test, commented on below).

azplugins/test-py/test_pair_hertz.py
mphoward: OK, no problem. You probably just missed one or two statements, but this will be fixed as part of a larger refactoring of the pair potential unit tests later.

azplugins/PairEvaluatorHertz.h
```diff
@@ -0,0 +1,106 @@
...
+ *      \f{eqnarray*}{
+ *      V(r)  = & \varepsilon (1-\frac{r}{r_{\mathrm{cut}}})^\frac{5}{2} & r < r_{\mathrm{cut}} \\
+ *            = & 0 r \ge r_{\mathrm{cut}}
+ *      \f}
+ *
+ * This potential is implemented as given in
+ * <a href="https://doi.org/10.1063/1.3186742">Pàmies, J. C., Cacciuto, A., & Frenkel, D., Journal of Chemical Physics, 131(4) (2009)</a>.
+ *
+ * The Hertz potential does not need diameter or charge. One parameter is specified and stored in a
+ * Scalar2.
```
mphoward, line 39:
```suggestion
 * Scalar.
```

azplugins/pair.py
```diff
@@ -199,6 +199,66 @@ def process_coeff(self, coeff):
 
         return _hoomd.make_scalar4(epsilon, sigma**3, sigma**6, _hoomd.int_as_scalar(style))
 
+class hertz(hoomd.md.pair.pair):
+    R""" Hertz potential
```
mphoward, line 203:
```suggestion
    R"""Hertz potential
```

azplugins/pair.py
```diff
@@ -199,6 +199,66 @@ def process_coeff(self, coeff):
...
+
+    :py:class:`hertz` is the Hertz potential
+    .. math::
+        :nowrap:
+
+        \begin{eqnarray*}{
+        V(r)  = & \varepsilon (1-\frac{r}{r_{\mathrm{cut}}})^{5/2} & r < r_{\mathrm{cut}} \\
+              = & 0 r \ge r_{\mathrm{cut}}
+        \f}
+        \end{eqnarray*}
```
mphoward, line 218: I think you still need to change this like so:
```suggestion
    :py:class:`hertz` is the Hertz potential:

    .. math::

        V(r)  &= \varepsilon \left(1-\frac{r}{r_{\rm cut}}\right)^{5/2},& r < r_{\rm cut} \\
              &= 0,& r \ge r_{\mathrm{cut}}
```

azplugins/pair.py
```diff
@@ -199,6 +199,66 @@ def process_coeff(self, coeff):
...
+        self.cpp_force = self.cpp_class(hoomd.context.current.system_definition, self.nlist.cpp_nlist, self.name)
+
+        hoomd.context.current.system.addCompute(self.cpp_force, self.force_name)
+
+        # setup the coefficent options
+        self.required_coeffs = ['epsilon']
+
+    def process_coeff(self, coeff):
+        epsilon = coeff['epsilon']
+        return epsilon
```
mphoward, line 260:
```suggestion
        return coeff['epsilon']
```

azplugins/test-py/test_pair_hertz.py
astatt: I've made some changes to the unit test, please have a look. The unit test was failing because after changing the import statement, the following code has to reference the classes explicitly, e.g.  ``` context. ``` --> ```hoomd.context``` etc. The easiest way to find those things is to run only one unit test with verbose output: ``` ctest -V -R azplugins-test_pair_hertz-cpu``` and then look from the top for error messages like: ```NameError: name 'context' is not defined``` and fix them one by one.

mphoward: Thank you Antonia!

mphoward:
Do you have an ETA for finishing up these changes? I'd like to merge all outstanding PRs by the end of the month and tag a new version.

arjungoswami:
Yes, sorry about that and thanks for letting me know. I'll look into this
ASAP (within the next couple of days) and if there's anything else you can
comment on it, so it can be finished by the end-of-month deadline.

Arjun

mphoward:
Thanks, I don't have any additional changes than the ones I already suggested.

mphoward:
Thank you for your work. I applied the small set of final changes to this branch so that I can finish another PR.

arjungoswami:
Thank you for finishing this up, and apologies that I was not able to do it
myself, as I was having some Internet connection issues on my end that were
just recently resolved. Hope everything is up-to-date now.

Issue #33, opened 2020-07-08
Remove from hoomd import * in unit tests


astatt:
For consistency, we should change and check the hoomd import statements in the *.py files, e.g. ```from hoomd import *``` to ``` import hoomd``` and associated necessary changes. This is not urgent by any means.

mphoward:
Yes, this is a good idea. The first pass solution would be to convert all the commands, prepending `hoomd` as needed.

I also would like to simplify the unit tests #34, which could make this easier. We can decide what order to do that in.

Issue #34, opened 2020-07-08
Condense unit tests


mphoward:
There is alot code duplication in unit tests, like those for the pair potentials. We want to do something like add a generic base class, which will have all of the tests implemented once. Deriving classes will specialize things like the name of the pair potential, the parameters to test, etc. This will make it easier to make changes in one place.

mphoward:
This has been completed as part of porting to HOOMD 4 and using pytest.

PR #36, opened 2020-07-25
Bugfix/hoomd import

astatt:
This bugfix addresses issue #33. It's pretty straightforward and hopefully I didn't overlook anything with my regex replacements. All tests still pass, so this should be ready to merge. Separately, we should address #34 at some point. 

I didn't think about it, should this go into develop or master?


mphoward:
Thank you very much! Please target this onto develop, and I will look it over and merge when I check on the other PRs. (Likely tomorrow night.)

mphoward:
Resolves #33

PR #37, opened 2020-08-24
Feature/bond double well

astatt:
This adds an additional parameter to the double well potential, such that the relative energy of the two minima can be tuned, i.e. by adding a linear term and tilting the potential.


azplugins/BondEvaluatorDoubleWell.h
```diff
@@ -31,32 +31,42 @@ namespace detail
...
+because of the added linear term.
+
 The parameters are:
-    - \a V_max (params.x) maximum potential value between the two minima
-    - \a a (params.y) shift for the location of the V_max, maximun is at a/2
-    - \a b (params.z) scaling for the distance of the two minima at (a/2 +/- b)
+    - \a V_max (params.x) potential difference between the the first minima and maxima
+    - \a a (params.y) shift for the location of the V_max, maximun is at approx a/2
+    - \a b (params.z) scaling for the distance of the two minima at approx (a/2 +/- b)
+    - \a b (params.w) potential difference between the two minima
```
mphoward, line 53:
```suggestion
    - \a c (params.w) potential difference between the two minima
```

azplugins/BondEvaluatorDoubleWell.h
```diff
@@ -31,32 +31,42 @@ namespace detail
...
+If c is non zero, the relative energy of the minima can be tuned, where c is the energy of the second minima,
+the first minima value is at zero. This  causes a small shift in the location of the minima and the maxima,
+because of the added linear term.
+
 The parameters are:
-    - \a V_max (params.x) maximum potential value between the two minima
-    - \a a (params.y) shift for the location of the V_max, maximun is at a/2
-    - \a b (params.z) scaling for the distance of the two minima at (a/2 +/- b)
+    - \a V_max (params.x) potential difference between the the first minima and maxima
+    - \a a (params.y) shift for the location of the V_max, maximun is at approx a/2
```
mphoward, line 51:
```suggestion
    - \a a (params.y) shift for the location of the V_max, maximum is at approx a/2
```

azplugins/bond.py
```diff
@@ -55,17 +55,22 @@ def __init__(self, name=None):
         hoomd.context.current.system.addCompute(self.cpp_force, self.force_name)
 
         # setup the coefficient options
-        self.required_coeffs = ['V_max','a','b']
+
+        self.required_coeffs = ['V_max','a','b','c']
```
mphoward, line 59: Does the HOOMD bond potential support default coefficients, like the pair potentials? If so, we should make `c=0` the default so that the original behavior of the class is maintained, and mark it in the documentation above.

astatt: It does! I didn't realize, this is a very good suggestion.

PR #38, opened 2020-08-27
Tutorials

astatt:
Three tutorials are written up to now: "Check hoomd-blue and azplugins installation", this one could/should also be going into "installation" (or developer topics) and then removed from the tutorials section. Migration would be easy, since it is only markdown text. Then there is a tutorial for "Reverse non-equilibrium simulation", both for LJ fluid and mpcd. The last tutorial is "Evaporation", both implicit (only thin film geometry, not spherical/droplet evaporation) and explicit. 

[Nbsphinx](https://nbsphinx.readthedocs.io/en/0.7.1/)  is needed to compile the Jupyter notebooks into the documentation.


mphoward:
These are looking overall very good! I have a few general comments:

1. I think we should remove Tutorial 0 and move it to the installation page. It is probably more appropriate there since it doesn't have very much interactive code / output. For now, you could just paste the text, and I can work on it directly.

2. Now that I see them in detail, I think that the reverse perturbation MD & MPCD pages should be separated into two different tutorials. You can refer to the MD one for some of the details of the setup, etc., but there are enough new equations etc. in the MPCD one that I think this is better treated as a separate topic.

3. The evaporation tutorials are good, but could we maybe find a way to condense some of this? I feel like there is a lot of boilerplate for setup, so maybe we could just supply small GSD files as starting configs? I also think it would be sufficient to have only one solute to reduce the number of parameters that need to be set throughout. Last, can we simplify the analysis in some way? Maybe we could use the FlowProfiler to get the density / temperature profiles? (If it doesn't supply temperature now, let's add that since it is easy.)

astatt:
- [ ] Move 'check installation'
- [ ] separate reverse perturbation tutorials in MD and MPCD part
- [ ] add temperature measurement to FlowProfiler 
- [ ] simplify setup of evaporation tutorials 
- [ ] simplify analysis using new FlowProfiler

mphoward:
@astatt I have had people reading these tutorials already and finding them helpful. Do you have anything outstanding on these? Otherwise, I will merge this branch and call it good enough for now.

astatt:
Let's finish and merge PR #40, then I can simplify the tutorials using the FlowProfiler (can do this over winter break - it will either get done very soon or not at all). I am also interested in keep adding tutorials for the other pieces in azplugins over time, but that's more of a long term thing and they can get their own branch when I have time to do this.

mphoward:
I merged up `develop` into this branch. When you get a chance, update the examples and we can call this a wrap.

Looking over the files, I know we had originally talked about keeping both Jupyter notebooks and python scripts. To simplify maintenance (updates like this), I would favor removing the python scripts and only using the Jupyter versions.

astatt:
I've updated the tutorial scripts. They are working and are useable. There are two potential things left to do: 

1. Simplify setup of all tutorials by using gsd files instead of setting up everything from scratch. 
2. Simplify the analysis for the evaporation. Using the FlowProfiler for it is possible, but we'd need to write the data and then reset it every time a measurement was made and we'd also need to extend it to gather data for different particle types. 

Any opinions on this?

mphoward:
> 1. Simplify setup of all tutorials by using gsd files instead of setting up everything from scratch.

Yes, I think this is a good idea for the simulations that use MD particles. Try to make the configurations as small as is reasonable to get decent results. It is OK to do the MPCD setup in the notebook since these are simple particles.

> 2. Simplify the analysis for the evaporation. Using the FlowProfiler for it is possible, but we'd need to write the data and then reset it every time a measurement was made and we'd also need to extend it to gather data for different particle types.

I don't want to try to support type filtering in the FlowProfiler at this time. I think we should simplify the example a bit. Use only one solute type, and use `numpy.histogram` to get the density profiles maybe at the beginning and end only?

astatt:
I've added gsd files for starting configurations for all MD simulations, simplified everything, pulled the analysis into post-processing of the trajectory, and divided the evaporation tutorial into two - explicit and implicit.  Hopefully I've addressed everything, let me know if I missed something.

mphoward (approved):
This is awesome! I quickly looked through the tutorials, and I think they all look good. We could eventually clean them up a little (for spelling, formatting, etc.), but I am happy with the content in them. I vote to merge now.

PR #40, opened 2020-10-05
Add temperature profile to FlowProfiler

astatt:
Small addition to the FlowProfiler to also be able to measure the temperature profile. 

Not sure if this is the easiest way of doing it, also it does not take any shear flow gradients into account. Not sure that we want to take flow gradients into account, because they are highly specific to each single potential use case, probably better to leave this simple let the user take care of those cases individually?  Otherwise, pretty straightforward.

Should be merged before the tutorials pull request, as it will be needed in evaporation tutorial.


azplugins/flow.py
```diff
@@ -757,3 +776,15 @@ def velocity(self):
             return np.divide(self._velocity, self._counts, out=np.zeros(self.bins), where=self._counts > 0)/self.samples
         else:
             return np.zeros(self.bins)
+
+    @property
+    def temperature(self):
```
mphoward, line 781: For consistency with the rest of HOOMD, let's call this property `kT` throughout because it is technically the thermal energy (Boltzmann's constant nominally being 1), even though this is essentially temperature.

azplugins/flow.py
```diff
@@ -726,13 +730,28 @@ def __call__(self, timestep):
             _velocity,_ = np.histogram(x, bins=self.bins, range=self.range, weights=v)
             self._velocity += _velocity
 
+            # temperature histogram
```
mphoward, line 733: To get the temperature, it might be better to do mass-averaged histograms? Also, since we are doing a lot of histograms, maybe we should combine all of these using `numpy.digitize` and `numpy.bincount`? I think something like this would cover all properties of interest:

```py
# add self.edges in constructor
binids = np.digitize(x, self.edges)

counts = np.bincount(binids)
bin_mass = np.bincount(binids, weights=m)
vn = np.bincount(binids, weights=v)
vm = np.bincount(binids, weights=v*m)
vsqm = np.bincount(binids, weights=v**2*m)

kT = vsqm/(3*(counts-1))
```

We could also clarify properties like `number_density` (from `counts`) vs. a new property `mass_density` (from `bin_mass`). We could also clarify that `velocity` is the number average, and add a new `mass_velocity` or something like that.

For `mass_density`, `mass_velocity`, and `temperature`, would you want to compute these in each frame (and then average over frames), or accumulate between frames and then average at the end? I think both are valid.

mphoward (changes requested):
I had a suggestion for computing the temperature & simplifying code while you are working on it. Let me know if you think this is reasonable or would work!

azplugins/flow.py
astatt: I would accumulate between frames and average when the user requests the data, but it makes no difference, this is just a preference.

mphoward: That's fine by me. I was worried about `<x/y>` vs. `<x>/<y>`, as I suppose this could make a difference if the number of counts in a bin is small.

astatt: The way you calculate the temperature now only takes the velocity component in `flow_axis` into account. I think I'd prefer to calculate the temperature from all velocity components, it also looks like that this is what you intended, by dividing by 3?

mphoward: Ah yes, that's a good point. Yes, it would make sense to use all three components to get kT. Do you want to assume that the flow is a vector (*u*(x)) or that it can only have one nonzero component (uy(x))?

If you assume a vector, we could modify the `velocity` so that it returns the vector (no need to specify `flow_axis`). `np.bincount` will not take a vector as a weight, but you could call it on each component of the velocity separately:
```py
m = snap.particles.mass
v = snap.particles.velocity
for dim in range(3):
   self._velocity[dim] += np.bincount(binids,v[:,dim])
   self._mass_velocity[dim] += np.bincount(binids,m*v[:,dim])
```

mphoward:
A test version of HOOMD 3.0 is now available. I'd like to merge this PR into `develop`, merge `develop` into `master`, and tag version `0.11.0`. I propose that this should be the last minor version tag that we add for HOOMD 2.x, and we work towards making the tutorials & additional features compatible with HOOMD 3.x in a new branch based off this.

How does this plan sound to you?

astatt:
Sounds good! I am going to be able to finish this up later today hopefully. For ''hoomd-3.x'', we should flip our major version number to ''1.x''.  "Open" features being worked on are: SLLOD, Dynamic bonding, and sine/cosine channels for mpcd. They should all transition to ''hoomd 3.x''.

mphoward:
Great, and agreed! Let me know if you need any help with this. I started to browse the new hoomd code, and it is a bit hard to follow because it is a preliminary release but I think we can start by getting things to compile and then start with the pair potentials. I will open a new issue to track this.

mphoward:
Also, I found out on another repo that because of the addition of the docker hub credentials, your PR may not automatically run through CircleCI anymore (security reasons). I can manually trigger the tests though when you request a review.

astatt:
``<x>/<y>`` vs ``<x/y>`` does make a difference for  extremely dilute systems. The latter will lead to many ``0/0``, which we set to ``0`` with ``np.divide()``. When averaged, all those empty ``0`` bin-counts do contribute to the average.  For a "dense" system, e.g a minimum of a particle in each bin, it makes no visible difference. 

I'm comparing: 

```
# measuring
for dim in range(3):
    self._number_velocity[dim] += np.bincount(binids,v[:,dim],minlength=self.bins)
# output
result = np.zeros_like(self._number_velocity)
if self.samples > 0:
   for dim in range(3):
         result[dim] = np.divide(self._number_velocity[dim], self._counts, out=np.zeros(self.bins), where=self._counts > 0)
```
and 

```
# measuring 
for dim in range(3):
    vel = np.bincount(binids,v[:,dim],minlength=self.bins)
    counts = np.bincount(binids,minlength=self.bins)
    self._number_velocity[dim] += np.divide(vel, counts, out=np.zeros(self.bins), where=counts > 0)

# output 
 result = np.zeros_like(self._number_velocity)
 if self.samples > 0:
      for dim in range(3):
          result[dim] = self._number_velocity[dim]/self.samples
```

mphoward:
Ah, that makes sense and what I was wondering. I would vote for using `<x/y>` because this is the _instantenous_ value of these properties. For temperature, that is the one that would be used by a profile-unbiased thermostat, so I think it makes the most sense to log this. As you say, for a larger system with many samples, there should be essentially no difference between the two.

azplugins/flow.py
```diff
@@ -757,3 +776,15 @@ def velocity(self):
             return np.divide(self._velocity, self._counts, out=np.zeros(self.bins), where=self._counts > 0)/self.samples
```
mphoward, line 776: Noting here: this line is incorrect. `self._counts` is already taking care of the total number of samples, so this is reducing the velocity by one factor too many.

mphoward:
Happy new year! I'm following up on this now because I tried to use it and found a bug, which I think I made. :-)

Could you push whatever changes you have, and I will finish off this PR and fix the bug? I am motivated because I need to use this in the near future. Thanks!

mphoward:
Hey @astatt: one more checkup on this. Do you have any commits that you haven't pushed yet? I will finish this one off myself and merge, but wanted to see if you had anything lying around. I am updating all of my repos now, trying to get a clean slate.

azplugins/flow.py
```diff
@@ -717,25 +722,41 @@ def __call__(self, timestep):
 
         if hoomd.comm.get_rank() == 0:
             x = snap.particles.position[:,self.bin_axis]
-            v = snap.particles.velocity[:,self.flow_axis]
+            v = snap.particles.velocity
+            m = snap.particles.mass
 
-            _counts,_ = np.histogram(x, bins=self.bins, range=self.range)
-            self._counts += _counts
+            binids = np.digitize(x, self.edges[1:])
```
mphoward, line 728: This seems a little funny because it is discarding the particles in the first bin (and making them fall in some error case). As I read the documentation, I think we want to use all the edges with `right=False` (the default), check if any of the bins are 0 (which would be an error), then subtract 1 from the bin so that they are 0-indexed. Do you agree?

azplugins/flow.py
```diff
@@ -717,25 +722,41 @@ def __call__(self, timestep):
...
 
-            _velocity,_ = np.histogram(x, bins=self.bins, range=self.range, weights=v)
-            self._velocity += _velocity
+            for dim in range(3):
+               num_vel = np.bincount(binids,v[:,dim],minlength=self.bins)
+               counts = np.bincount(binids,minlength=self.bins)
+               self._mass_velocity[dim] += np.bincount(binids,m*v[:,dim],minlength=self.bins)
+               self._number_velocity[dim] += np.divide(num_vel, counts, out=np.zeros(self.bins), where=counts > 0)
```
mphoward, line 734: I think you want to do something a little more like this:
```suggestion
            counts = np.bincount(binids,minlength=self.bins)
            self._counts += counts

            mass = np.bincount(binids,weight=m,minlength=self.bins)
            self._bin_mass += mass

            num_vel = np.zeros((self.bins,3))
            mass_vel = np.zeros((self.bins,3))
            for dim in range(3):
               num_vel[:,dim] = np.bincount(binids,v[:,dim],minlength=self.bins)
               mass_vel[:,dim] = np.bincount(binids,m*v[:,dim],minlength=self.bins)
           np.divide(num_vel, counts, out=num_vel, where=counts > 0)
           np.divide(mass_vel, mass, out=mass_vel, where=mass > 0)
           self._number_velocity[dim] += num_vel
           self._mass_velocity[dim] += mass_vel
```

azplugins/flow.py
```diff
@@ -717,25 +722,41 @@ def __call__(self, timestep):
...
 
+            vsq = np.sum(v**2,axis=1)
+            vsqm = np.bincount(binids, weights=vsq*m, minlength=self.bins)
+            self._vsqm +=vsqm
+
+            self._counts += np.bincount(binids, minlength=self.bins)
+            self._bin_mass += np.bincount(binids, weights=m, minlength=self.bins)
```
mphoward, line 741: These aren't needed if you do this calculation above
```suggestion
```

azplugins/flow.py
```diff
@@ -717,25 +722,41 @@ def __call__(self, timestep):
...
-            self._velocity += _velocity
+            for dim in range(3):
+               num_vel = np.bincount(binids,v[:,dim],minlength=self.bins)
+               counts = np.bincount(binids,minlength=self.bins)
+               self._mass_velocity[dim] += np.bincount(binids,m*v[:,dim],minlength=self.bins)
+               self._number_velocity[dim] += np.divide(num_vel, counts, out=np.zeros(self.bins), where=counts > 0)
 
+            vsq = np.sum(v**2,axis=1)
+            vsqm = np.bincount(binids, weights=vsq*m, minlength=self.bins)
+            self._vsqm +=vsqm
```
mphoward, line 738: Now for this calculation, we can subtract the center-of-mass velocity from `mass_vel`. I think there's a way to do this just correcting the kinetic energy of the particles (without this velocity subtracted). The formulas are in the MPCD temperature compute code... I can look for them if you want me to.

azplugins/flow.py
```diff
@@ -746,13 +767,51 @@ def density(self):
...
-    def velocity(self):
-        r"""The current average velocity profile."""
+    def mass_density(self):
+        r"""The current mass-averaged density profile."""
+        if hoomd.comm.get_rank() != 0:
+            hoomd.context.msg.error('Flow profile only defined on root rank.\n')
+            raise RuntimeError('Flow profile only defined on root rank')
+
+        if self.samples > 0:
+            return np.divide(self._bin_mass, self._counts, out=np.zeros(self.bins), where=self._counts > 0)
```
mphoward, line 777: With the rewrite I did above, I think you just divide by the volume and samples, like for the number density,

azplugins/flow.py
```diff
@@ -746,13 +767,51 @@ def density(self):
...
+    def number_velocity(self):
+        r"""The current number-averaged velocity profile ."""
+        if hoomd.comm.get_rank() != 0:
+            hoomd.context.msg.error('Flow profile only defined on root rank.\n')
+            raise RuntimeError('Flow profile only defined on root rank')
+
+        result = np.zeros_like(self._number_velocity)
+        if self.samples > 0:
+            for dim in range(3):
+               result[dim] = self._number_velocity[dim]/self.samples
```
mphoward, line 791: Can't you just do:
```
if self.samples > 0:
    return self._number_velocity / self.samples
else:
    return np.zeros(self.bins)
```
?

azplugins/flow.py
```diff
@@ -746,13 +767,51 @@ def density(self):
...
+        r"""The current mass-averaged velocity profile ."""
+        if hoomd.comm.get_rank() != 0:
+            hoomd.context.msg.error('Flow profile only defined on root rank.\n')
+            raise RuntimeError('Flow profile only defined on root rank')
+
+        result = np.zeros_like(self._mass_velocity)
+        if self.samples > 0:
+            for dim in range(3):
+               result[dim] = np.divide(self._mass_velocity[dim], self._counts, out=np.zeros(self.bins), where=self._counts > 0)
+        return result
```
mphoward, line 805: Same thing here.

azplugins/flow.py
```diff
@@ -746,13 +767,51 @@ def density(self):
...
+    @property
+    def kT(self):
+        r"""The current average temperature profile."""
         if hoomd.comm.get_rank() != 0:
             hoomd.context.msg.error('Flow profile only defined on root rank.\n')
             raise RuntimeError('Flow profile only defined on root rank')
 
         if self.samples > 0:
-            return np.divide(self._velocity, self._counts, out=np.zeros(self.bins), where=self._counts > 0)
+            return np.divide(self._vsqm, (3*(self._counts-1)), out=np.zeros(self.bins), where=(3*(self._counts-1)) > 0)
```
mphoward, line 815: I think if we correct code above, this will also look like the velocity profile code, just divide the internal kT by samples.

mphoward (changes requested):
This is definitely going in the right direction! Here are some comments, and we can discuss further if you want.

Can you also please remove the jupyter notebooks from this branch? Thanks!

azplugins/flow.py
```diff
@@ -657,12 +657,13 @@ def set_params(self, group=None, Nswap=None, width=None, target_momentum=None):
...
     R"""Measure average profiles along a spatial dimension.
 
-    The average density and velocity profile is computed along a given spatial dimension.
+    The average density and velocity profile is computed along a given spatial dimension. Both
+    number-averaged quantities are available, `number_density` and `number_velocity`, as well as mass-averaged
+    profiles, `mass_density` and `mass_velocity`.
 
     Args:
         system: :py:mod:`hoomd` or :py:mod:`hoomd.mpcd` system (e.g., returned by :py:func:`hoomd.init.read_gsd`).
         bin_axis (int): direction for binning (0=`x`, 1=`y`, 2=`z`).
```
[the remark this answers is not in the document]
astatt: No preference between `str` vs `int`. Is there any hoomd precedent for naming convention we can follow? I will keep this as an `int` for now and rename it to `axis`.

azplugins/flow.py
```diff
@@ -673,31 +674,35 @@ class FlowProfiler:
...
+            np.savetxt('profiles.dat', np.column_stack((f.centers, f.number_density, f.number_velocity[:,2], f.kT)))
 
     .. note::
 
         In MPI simulations, the profiles are **only** valid on the root rank. Accessing them on
         other ranks will produce an error.
 
+    .. note::
+
+        The temperature profile `kT` is calculated from the velocity values without substracting any shear flow.
```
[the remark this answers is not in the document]
astatt: We discussed it briefly -- I remember a vaguely that we wanted to keep the flow profiler generic and applicable to different types of flow. But, we should be able to do that anyway by subtracting the average velocity in each bin and looking at the velocity variance, correct? Is information about the flow (direction, symmetry etc) needed for calculating kT?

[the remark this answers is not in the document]
mphoward: Yeah, I don’t remember, but the profile unbiased temperature makes the most sense to me. You don’t need any extra information about the flow, I think. You compute the kinetic energy directly from the velocities, then subtract the kinetic energy of the center of mass of the bin, like here:

https://github.com/glotzerlab/hoomd-blue/blob/ed60448e7720fe6bade26ca53876259d93eec9ea/hoomd/mpcd/CellThermoCompute.cc#L340

You already have all the information needed for the center-of-mass KE since you know the total mass and the mass-average velocity of the bin.

The temperature is zero if there are 0 or 1 particles in the cell.

azplugins/flow.py
astatt: Questions to consider:

- should we check that the range is inside of the box? 
- if the user specifies a range smaller than the box, particles could be outside of the range, so maybe that case should not be an error, but those particles should be filtered out/ignored?

azplugins/flow.py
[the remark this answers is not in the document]
mphoward: I'm not sure if there is a precedent, so let's leave it as an int for now.

azplugins/flow.py
astatt: yes, absolutely. This is simply leftover copy-pasted code from prior versions.

azplugins/flow.py
astatt: Please double check, I think the temperature is correct now.

azplugins/flow.py
astatt: for now: no checks if `range` is inside box, added comment to documentation.

mphoward (approved):
Everything looks good! I will merge once the CI tests finish running.

PR #47, opened 2021-01-11
Remove erroneous normalization of flow velocity by samples

mphoward:
Removes extra factor of `self.samples` when averaging velocities, which caused the reported velocities to be too low.


mphoward:
@astatt This is a hotfix, so if you could check and approve, I will merge asap.

mphoward:
Arash reviewed this and it is a small change, so I merged.

Issue #56, opened 2022-01-05
Revise documentation


mphoward:
This is a new issue (building from #1) to track edits and revisions that need to be made to the documentation:

- [ ] Main index should either embed the readme or have some text included
- [ ] Installation instructions need to be condensed (borrow from readme). The tutorial text that is there is for verifying an installation.
- [ ] Add module level documentation for pretty much all modules.
- [ ] Add overview text for Potentials index page (in addition to toc).
- [ ] Math equations are rather broken, try to clean them up.
- [ ] Check for use of single vs. double ticks throughout.
- [ ] Collapse release notes table of contents by adding a `v0.x` header. This will be useful when we finally get to `v1.0.0`.
- [ ] Check tutorials for typos. At least one of them is misnumbered since there are two "Tutorial 3". I would propose that we actually get these separated out into individual directories (do not include two related notebooks in one folder) to make it easier to count them.
- [ ] Proofread all modules.
- [ ] Remove test version of HOOMD from README. It is easy to let it get out of date.
- [ ] Add badge to README for RTD.
- [ ] Check for missing `:members:` documentation where it's needed.

Anyone should feel free to work on these. Please document on this thread if you complete an item, especially if you proofread a module.

astatt:
Do you want to address these points in a separate branch or as hotfixes?

mphoward:
They can be addressed whenever we feel like it, in any size chunk, and merged immediately to `main`. (I am going to remove the `develop` branch after we merge the last open PRs.)

mphoward:
The documentation will need to be redone when v1 is ready.

mphoward:
Documentation was rewritten for v1. Any changes and additions are welcome in new PRs.

PR #57, opened 2022-01-21
Use GitHub Actions for testing

### LINKED ISSUE #41: Remove unit tests for old HOOMD versions
HOOMD 2.x has become pretty stable, so I think we should remove tests using the old images to shorten testing time. I would propose that we build a new image with HOOMD 2.9.3 and target this as the last 2.x version that we will support before HOOMD 3.x. We could potentially run one or two tests on older versions if we want to continue to promise support for a minimum version of HOOMD.

### LINKED ISSUE #54: Switch unit tests to GitHub Actions
CircleCI is a pain and GitHub Actions is clean. We should migrate the tests to this platform, and simplify the configurations we test.

mphoward:
This PR enables testing with GitHub Actions. It uses the same test containers and settings as HOOMD:

https://github.com/glotzerlab/hoomd-blue/blob/master/.github/workflows/test.yml

This configuration limits execution to a handful of cases (rather than a full matrix). Some limitations are that we only do double-precision builds, all builds have both MD and MPCD components, and we continue to only test for compilation under CUDA (no execution). The advantages are that we can now uses the glotzerlab Docker containers so that we don't have to build our own, we shorten the number of tests, and we can test multiple versions of HOOMD more easily.

I have removed the code format checker from the unit tests. The script is not so robust, and this would be better handled with another tool like `pre-commit`.

Todo:
- [x] Should we try to use `actions/cache` to store compiled HOOMD to speedup some builds? This is helpful when making changes to azplugins, as HOOMD does not change. However, it means we have to be careful to invalidate the cache when the build environment changes.
- [x] Should we do internal builds, external+internal builds, or external builds only (current setup)?
- [x] Remove rest of circleci leftovers
- [x] Update required tests
- [x] Add some developer documentation on what is tested

Resolves #41
Resolves #54


astatt:
I think the proposed simplification of testing is appropriate. I would favor to simply do external build tests, but I am biased because that is how I tend to install azplugins. 

How big is the potential speedup by using `actions/cache`? At least in the workflow I normally use for coding, I do execute a sub-set of unit-tests periodically and locally, and really only rely on the build-in testing once I get close to a pull request. So, the speed has not really been a limiting factor for me. If it is difficult to make sure to  be careful about invalid caches, to me it is only worth it if the speedup is significant. 

I think single precision vs double is something we do not want to address as developers/maintainers, but each user has to think about this carefully themselves. So testing double only is fine in my opinion. We can revisit this when hoomd has a mixed precision mode, because then the potential performance gain on all the ML optimized GPUs might be worth it.  

We might want to consider writing a paragraph about how testing works in principle and what is tested/not tested, as well as additional recommended testing in the Developer Guidelines. It could also go onto the to do list #56. 

So overall, this is good! In terms of hoomd-versions to test: definitively  2.9.7 as last stable v2, and perhaps 2.6.0 or 2.8.0? I think we do not have to go much older versions.

mphoward:
Awesome, thanks for the feedback!

> I think the proposed simplification of testing is appropriate. I would favor to simply do external build tests, but I am biased because that is how I tend to install azplugins.

Sounds good, external only is easy.

> How big is the potential speedup by using `actions/cache`? At least in the workflow I normally use for coding, I do execute a sub-set of unit-tests periodically and locally, and really only rely on the build-in testing once I get close to a pull request. So, the speed has not really been a limiting factor for me. If it is difficult to make sure to be careful about invalid caches, to me it is only worth it if the speedup is significant.

It currently takes about 6 minutes to compile HOOMD, then 1.5 minutes to build and run azplugins. So, it is a large relative amount of time, but in absolute time it is not very much. Considering that we are using GitHub runners, we don't really need to worry about contesting resources, so I think we can just leave it be for now.

> I think single precision vs double is something we do not want to address as developers/maintainers, but each user has to think about this carefully themselves. So testing double only is fine in my opinion. We can revisit this when hoomd has a mixed precision mode, because then the potential performance gain on all the ML optimized GPUs might be worth it.

I agree. We might catch some compilation errors with single though, if someone accidentally codes the wrong precision? I think for now, we should not test this if HOOMD also does not test it, and we can address in a future PR if we support mixed precision.

> We might want to consider writing a paragraph about how testing works in principle and what is tested/not tested, as well as additional recommended testing in the Developer Guidelines. It could also go onto the to do list #56.

I can add this

> So overall, this is good! In terms of hoomd-versions to test: definitively 2.9.7 as last stable v2, and perhaps 2.6.0 or 2.8.0? I think we do not have to go much older versions.

Given that HOOMD v2 is not going to be supported post March, I don't think I want to worry about testing backwards compatibility v2.x. We are going to have to make a clean break to v3, so let's focus our efforts on testing multiple versions of that when they become available.

astatt (approved):
As soon as you have resolved the two remaining tasks to your satisfaction this can be merged! And the developer info for testing can be short for now, and extended over time if necessary.

PR #59, opened 2022-02-02
Add basic HOOMD version check python decorator

mphoward:
This branch has been floating around. Not sure if it is useful or not, so wanted to put it up as a PR and we can decide if we want to merge or close.

Related to #10

### DIFF
--- a/azplugins/api.py
+++ b/azplugins/api.py
@@ -1,0 +1,64 @@
+# Copyright (c) 2018-2019, Michael P. Howard
+# This file is part of the azplugins project, released under the Modified BSD License.
+
+# Maintainer: mphoward
+
+r"""Utility methods for interfacing with the HOOMD API."""
+import hoomd
+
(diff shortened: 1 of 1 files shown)


astatt (commented):
So, how much of this would translate well to hoomd v3? 

While this is in general useful, I am not sure that we are not creating a small utility that will be annoying to maintain in the future, that's my only concern.

Copyright is out of date.

mphoward:
When we move to v3, we will be making a clean break in azplugins (major version bump), so we would not need a maximum version check at that point. I think the purpose of this decorator was to stop someone from trying to use a feature in azplugins that was not supported in an older version of HOOMD, but we could also extend it to enforce a max version as well (e.g., if a feature gets deprecated).

I honestly forget why I even wrote this in the first place. My feeling is that it is not going to be that useful, so I would vote to close the PR.

PR #63, opened 2022-03-04
Allow epsilon and/or sigma to be zero in FENE bonds

judevishnu:
Hi Mike,
I made some the requested changes to code for allowing epsilon=0 parameters. Please verify.

Regards,
Jude

### DIFF
--- a/azplugins/BondEvaluatorFENE.h
+++ b/azplugins/BondEvaluatorFENE.h
@@ -108,7 +108,7 @@ class BondEvaluatorFENE
         DEVICE bool evalForceAndEnergy(Scalar& force_divr, Scalar& bond_eng)
             {
             // check for invalid parameters
-            if (r_0 == Scalar(0.0) ) return false;
+            if (r_0 == Scalar(0.0)) return false;
 
             // Check if bond length restriction is violated
             if (rsq >= r_0*r_0) return false;
@@ -118,7 +118,7 @@ class BondEvaluatorFENE
             Scalar sigma6inv = lj2/lj1;
 
             // wca cutoff: r < 2^(1/6)*sigma
-            //If epsion or sigma is zero or beyond cutoff force and energy are zero
+            // if epsilon or sigma is zero OR r is beyond cutoff, the force and energy are zero
             if (lj1 != Scalar(0) && r6inv > sigma6inv/Scalar(2.0))
                 {
                 Scalar epsilon = lj2*lj2/Scalar(4.0)/lj1;
@@ -127,8 +127,8 @@ class BondEvaluatorFENE
                 }
             else
                 {
-                force_divr=0;
-                bond_eng=0;
+                force_divr = 0;
+                bond_eng = 0;
                 }
             // Check if delta is non -zero,if non-zero sqrt of rsq is calculated
             if (delta != Scalar(0))
--- a/azplugins/BondEvaluatorFENE.h
+++ b/azplugins/BondEvaluatorFENE.h
@@ -108,7 +108,7 @@ class BondEvaluatorFENE
         DEVICE bool evalForceAndEnergy(Scalar& force_divr, Scalar& bond_eng)
             {
             // check for invalid parameters
-            if (lj1 == Scalar(0.0) || r_0 == Scalar(0.0) || K == Scalar(0.0) ) return false;
+            if (r_0 == Scalar(0.0) || K == Scalar(0.0) ) return false;
 
             // Check if bond length restriction is violated
             if (rsq >= r_0*r_0) return false;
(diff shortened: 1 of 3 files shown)


azplugins/BondEvaluatorFENE.h
```diff
@@ -108,7 +108,7 @@ class BondEvaluatorFENE
         DEVICE bool evalForceAndEnergy(Scalar& force_divr, Scalar& bond_eng)
             {
             // check for invalid parameters
-            if (lj1 == Scalar(0.0) || r_0 == Scalar(0.0) || K == Scalar(0.0) ) return false;
+            if (r_0 == Scalar(0.0) || K == Scalar(0.0) ) return false;
```
mphoward, line 111: This is right, but I am actually not sure why we are validating parameters inside the kernel here. Let's leave the check on `r_0` since it prevents divide-by-zero, but I don't see a problem with `K == 0` either:
```suggestion
            if (r_0 == Scalar(0.0)) return false;
```

azplugins/BondEvaluatorFENE.h
```diff
@@ -119,12 +119,18 @@ class BondEvaluatorFENE
...
+            if (lj1 != Scalar(0) && r6inv > sigma6inv/Scalar(2.0))
                 {
                 Scalar epsilon = lj2*lj2/Scalar(4.0)/lj1;
                 force_divr = r2inv * r6inv * (Scalar(12.0)*lj1*r6inv - Scalar(6.0)*lj2);
                 bond_eng = r6inv * (lj1*r6inv - lj2) + epsilon;
                 }
+            else
+                {
+                bond_eng=0;
+                force_divr=0;
```
mphoward, line 132:
```suggestion
                force_divr = 0;
                bond_eng = 0;
```

azplugins/BondEvaluatorFENE.h
```diff
@@ -119,12 +119,18 @@ class BondEvaluatorFENE
 
             // wca cutoff: r < 2^(1/6)*sigma
             // wca cutoff: 1 / r^6 > 1/((2^(1/6))^6 sigma^6)
-            if (r6inv > sigma6inv/Scalar(2.0))
+            //If epsion is zero force and energy is zero
```
mphoward, line 122:
```suggestion
            // if epsilon or sigma is zero OR beyond cutoff, force and energy are zero
```

azplugins/bond.py
```diff
@@ -161,12 +161,6 @@ def process_coeff(self, coeff):
         lj1 = 4.0 * epsilon * math.pow(sigma, 12.0)
         lj2 = 4.0 * epsilon * math.pow(sigma, 6.0)
 
-        if epsilon==0:
-            hoomd.context.msg.error("azplugins.bond.fene(): epsilon must be non-zero.\n")
-            raise ValueError('epsilon must be non-zero')
-        if sigma==0:
-            hoomd.context.msg.error("azplugins.bond.fene(): sigma must be non-zero.\n")
-            raise ValueError('sigma must be non-zero')
         if k==0:
```
mphoward, line 164: Suggest to remove this too, it is not an error to have k = 0 as far as I know, although it would be silly.

azplugins/test-py/test_bond_fene.py
```diff
@@ -52,16 +52,6 @@ def test_set_missing_r0(self):
...
-        fene = azplugins.bond.fene()
-        fene.bond_coeff.set('bond', epsilon=0, sigma=1.0, k=30,r0=1.5)
-        self.assertRaises(ValueError, fene.update_coeffs)
-
-    def test_set_zero_sigma(self):
-        fene = azplugins.bond.fene()
-        fene.bond_coeff.set('bond', epsilon=1.0, sigma=0.0, k=30, r0=1.5)
-        self.assertRaises(ValueError, fene.update_coeffs)
-
     def test_set_zero_k(self):
```
mphoward, line 55: Suggest to remove (see above)

mphoward (changes requested):
This looks good! I have a request for you to also remove the check on K != 0 in this same PR. Otherwise, I think this will work just fine if all tests pass.

azplugins/test-py/test_bond_fene.py
judevishnu: Should I add a test when k=0?

mphoward: It is not necessary. Unlike for the LJ parameters, there is no place where K could be used to divide by zero.

judevishnu:
Hi Mike,
I think I have made all your requested changes. Let me know if somethings not needed or need to be added.
Thanks,
Jude

azplugins/test-py/test_bond_fene.py
judevishnu: Oh, I did not see you message and added it and asked for a review. But no worries, I will remove it

azplugins/test-py/test_bond_fene.py
```diff
@@ -144,6 +129,202 @@ def test_potential_delta_nonzero(self):
...
+        self.assertAlmostEqual(f0[0],-F,3)
+        self.assertAlmostEqual(f0[1],0)
+        self.assertAlmostEqual(f0[2],0)
+
+        self.assertAlmostEqual(f1[0],F,3)
+        self.assertAlmostEqual(f1[1],0)
+        self.assertAlmostEqual(f1[2],0)
+
+    # test the calculation of foce and potential when k=0 and epsilon=nonzero sigma=nonzero
+    def test_potential_k_zero_epsilon_nonzero_sigma_nonzero(self):
```
mphoward, line 217:
```suggestion
    def test_potential_k_zero(self):
```

azplugins/test-py/test_bond_fene.py
```diff
@@ -144,6 +129,202 @@ def test_potential_delta_nonzero(self):
...
+        self.assertAlmostEqual(e1,0.5*U,3)
+
+        self.assertAlmostEqual(f0[0],-F,3)
+        self.assertAlmostEqual(f0[1],0)
+        self.assertAlmostEqual(f0[2],0)
+
+        self.assertAlmostEqual(f1[0],F,3)
+        self.assertAlmostEqual(f1[1],0)
+        self.assertAlmostEqual(f1[2],0)
+
```
mphoward, line 327:
```suggestion
```

mphoward (changes requested):
Thank you! I am going to batch commit the suggestions I put on this PR and merge, so no further work is needed on your end.

mphoward (approved):

PR #64, opened 2022-03-04
Compute center of mass velocity of a group of particles

mphoward:
This PR implements a CPU compute to add up the velocity of a group of particles. This is useful in MPI simulations because it bypasses needing to pull a snapshot onto the root rank in python. It could be implemented on the GPU too using `thrust::transform_reduce`, but I don't have a GPU for testing and don't need it right now.

### DIFF
--- a/azplugins/GroupVelocityCompute.cc
+++ b/azplugins/GroupVelocityCompute.cc
@@ -23,7 +23,11 @@ GroupVelocityCompute::GroupVelocityCompute(std::shared_ptr<SystemDefinition> sys
     : Compute(sysdef),
       m_group(group),
       m_lognames{"vx"+suffix,"vy"+suffix,"vz"+suffix},
-      m_velocity(0,0,0)
+      m_velocity(make_scalar3(0,0,0))
+    {
+    }
+
+GroupVelocityCompute::~GroupVelocityCompute()
     {
     }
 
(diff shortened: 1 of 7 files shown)


mphoward:
I confirmed that this analyzer produces the same results as one written in Python on stampede2. @astatt could you please review when you get a chance so I can merge?

azplugins/GroupVelocityCompute.cc
```diff
@@ -0,0 +1,118 @@
...
+        {
+        m_velocity = make_scalar3(0,0,0);
+        return;
+        }
+
+    // accumulate momentum and masses
+    const unsigned int N = m_group->getNumMembers();
+    ArrayHandle<unsigned int> h_index(m_group->getIndexArray(), access_location::host, access_mode::read);
+    ArrayHandle<Scalar4> h_vel(m_pdata->getVelocities(), access_location::host, access_mode::read);
+    Scalar3 momentum = make_scalar3(0.,0.,0.);
```
astatt, line 57: same comment as other PR

azplugins/analyze.py
```diff
@@ -18,6 +21,58 @@
...
+    ``vy``, and ``vz`` with ``suffix`` appended. By default, ``suffix`` is
+    ``_name`` where ``name`` is the name of the ``group``, like ``_all`` for
+    :py:class:`hoomd.group.all`. However, a custom suffix may be specified; the
+    only requirement is that the same suffix cannot be used more than once. You
+    can save these results using :py:class:`hoomd.analyze.log`.
+
+    Example::
+
+        azplugins.analyze.group_velocity(hoomd.group.all())
+        hoomd.analze.log(filename='velocity.dat', quantities=['vx_all'], period=10)
```
astatt, line 50: same as other PR

astatt (approved):
Identical comments to the other PR. I should have looked at them in the opposite order I suppose.  Anyway, ready to merge.

azplugins/analyze.py
mphoward: Typo fixed and two more examples added.

mphoward:
> If we every need the GPU component, I can add it and test it.

Sounds good! Arash's student could also implement this if he wants to use this analyzer on their GPUs. I wrote it because of how slow the Python analyzer was on Stampede2, but they see similar behavior due to host-device copy bottleneck.

PR #73, opened 2024-06-03
Refactor Ashbaugh-Hatch potential

mphoward:
This PR refactors `azplugins.pair.ashbaugh` into `azplugins.pair.PerturbedLennardJones`. Compared to v0.12.0, the main changes are to rename the class and remove the alpha parameter, which was rarely (if ever) used.

This PR is WIP since the class needs documentation and other test cases need to be ported.


src/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ set(python_files
     pair.py
     )
 
-set(_pair_evaluators )
+set(_pair_evaluators PerturbedLennardJones)
```
mphoward, line 20: Add your new evaluator to the list here. The name of the evaluator and the Python class should be the same.
```suggestion
set(_pair_evaluators PerturbedLennardJones)
```

src/PairEvaluatorPerturbedLennardJones.h
```diff
@@ -0,0 +1,176 @@
+// Copyright (c) 2018-2020, Michael P. Howard
+// Copyright (c) 2021-2024, Auburn University
+// Part of azplugins, released under the BSD 3-Clause License.
+
+#ifndef AZPLUGINS_PAIR_EVALUATOR_PERTURBED_LENNARD_JONES_H_
```
mphoward, line 5: Include guards should start with `AZPLUGINS_` then be the name of the file.

src/PairEvaluatorPerturbedLennardJones.h
```diff
@@ -0,0 +1,176 @@
...
+#endif
+
+namespace hoomd
+    {
+namespace azplugins
+    {
+namespace detail
+    {
+//! Define the parameter type used by this pair potential evaluator
+struct PairParametersPerturbedLennardJones : public PairParameters
```
mphoward, line 25: This struct implements the conversion between Python and C++. You need to have a default constructor (no arguments) and one that takes a Python dictionary & bool, as below.

src/PairEvaluatorPerturbedLennardJones.h
```diff
@@ -0,0 +1,176 @@
...
+        const Scalar sigma_2 = sigma * sigma;
+        const Scalar sigma_4 = sigma_2 * sigma_2;
+        sigma_6 = sigma_2 * sigma_4;
+
+        epsilon_x_4 = Scalar(4.0) * epsilon;
+        lam = v["lam"].cast<Scalar>();
+        rwcasq = pow(Scalar(2.), 1. / 3.) * sigma_2;
+        }
+
+    pybind11::dict asDict()
```
mphoward, line 44: This method inverts the constructor above.

src/PairEvaluatorPerturbedLennardJones.h
```diff
@@ -0,0 +1,176 @@
...
+    Scalar sigma_6;     //!< The coefficient for 1/r^12
+    Scalar epsilon_x_4; //!< The coefficient for 1/r^6
+    Scalar lam;         //!< Controls the attractive tail, between 0 and 1
+    Scalar rwcasq;      //!< The square of the location of the LJ potential minimum
+    }
+#if HOOMD_LONGREAL_SIZE == 32
+    __attribute__((aligned(16)));
+#else
+    __attribute__((aligned(32)));
+#endif
```
mphoward, line 63: Alignment should be a power of two and based on the size of the data. Usually 8, 16, or 32 bytes is appropriate.

src/PairEvaluatorPerturbedLennardJones.h
```diff
@@ -0,0 +1,176 @@
...
+ *      \f}
+ *
+ * where \f$V_{\mathrm{LJ}}(r,\varepsilon,\sigma)\f$ is the standard
+ * Lennard-Jones potential (see EvaluatorPairLJ) with parameters
+ * \f$\varepsilon\f$ and \f$\sigma\f$.
+ */
+class PairEvaluatorPerturbedLennardJones : public PairEvaluator
+    {
+    public:
+    typedef PairParametersPerturbedLennardJones param_type;
```
mphoward, line 83: This needs to match the parameter class you make above.

src/PairEvaluatorPerturbedLennardJones.h
```diff
@@ -0,0 +1,176 @@
...
+    //! Constructor
+    /*!
+     * \param _rsq Squared distance between particles
+     * \param _rcutsq Cutoff radius squared
+     * \param _params Pair potential parameters, given by typedef above
+     *
+     * The functor initializes its members from \a _params.
+     */
+    DEVICE
+    PairEvaluatorPerturbedLennardJones(Scalar _rsq, Scalar _rcutsq, const param_type& _params)
```
mphoward, line 94: This calls the parent constructor with the first two arguments, then unpacks the rest into any internal form of the parameters.

src/module.cc
```diff
@@ -49,6 +49,13 @@ namespace azplugins
     {
 namespace detail
     {
+
+void export_PotentialPairPerturbedLennardJones(pybind11::module&);
```
mphoward, line 53: Add forward declaration of the export method for your potential alphabetically to both of these lists. The first is the CPU version, the second is the GPU version. The code that implements this will be automatically generated by CMake.

src/module.cc
```diff
@@ -57,4 +64,10 @@ namespace detail
 PYBIND11_MODULE(_azplugins, m)
     {
     using namespace hoomd::azplugins::detail;
+
+    export_PotentialPairPerturbedLennardJones(m);
+
+#ifdef ENABLE_HIP
+    export_PotentialPairPerturbedLennardJonesGPU(m);
+#endif // ENABLE_HIP
```
mphoward, line 72: Then also call these methods here.

src/pair.py
```diff
@@ -3,3 +3,25 @@
...
 
 """Pair potentials."""
+
+from hoomd.azplugins import _azplugins
+from hoomd.data.parameterdicts import TypeParameterDict
+from hoomd.data.typeparam import TypeParameter
+from hoomd.md import pair
+
+
+class PerturbedLennardJones(pair.Pair):
```
mphoward, line 13: Add your potential by adapting this class. Classes should usually be added alphabetically.

src/pair.py
```diff
@@ -3,3 +3,25 @@
...
+    _ext_module = _azplugins
+    _cpp_class_name = 'PotentialPairPerturbedLennardJones'
+    _accepted_modes = ('none', 'shift', 'xplor')
+
+    def __init__(self, nlist, default_r_cut=None, default_r_on=0, mode='none'):
+        super().__init__(nlist, default_r_cut, default_r_on, mode)
+        params = TypeParameter(
+            'params',
+            'particle_types',
+            TypeParameterDict(epsilon=float, sigma=float, lam=float, len_keys=2),
```
mphoward, line 25: This is the part you need to change to match the parameters of your potential. `len_keys` stays equal to 2, it refers to the particle types.

src/pytest/test_pair.py
```diff
@@ -15,7 +15,17 @@
     ['potential', 'params', 'r_cut', 'shift', 'distance', 'energy', 'force'],
 )
 
-potential_tests = []
+potential_tests = [
+    PotentialTestCase(
```
mphoward, line 19: Add test cases for your potential here that you have computed by hand, with Mathematica or Python, etc. The forces and energies need to be recorded to 4 decimal places. See the named tuple above for an explanation of the meaning of each of these. You can (should) add multiple test cases for each of your potentials.

astatt: Also, add a short comment explaining the reasoning for each test, i.e "testing that potential and force are zero outside of cutoff" or "testing that if epsilon is zero, the potential is zero" etc.

src/pair.py
```diff
@@ -66,3 +66,76 @@ def __init__(self, nlist, default_r_cut=None, default_r_on=0, mode='none'):
...
+        The Perturbed LJ potential parameters. The dictionary has the following
+        keys:
+
+        * ``epsilon`` (`float`, **required**) - energy parameter
+          :math:`\varepsilon` :math:`[\mathrm{energy}]`
+        * ``sigma`` (`float`, **required**) - particle size :math:`\sigma`
+          :math:`[\mathrm{length}]`
+        * ``lam`` (`float`, **required**) - scale factor for attraction,
+        between 0 and 1 :math:`\lambda`
+          :math:`[\mathrm{dimensionless}]`
```
mphoward, line 118: @astatt Should we rename this to something more descriptive? It is `lam` because `lambda` is a reserved keyword in Python. Maybe we should call it `attraction_scale_factor`?
```suggestion
        * ``attraction_scale_factor`` (`float`, **required**) - scale factor for attraction :math:`\lambda`,
          usually between 0 and 1 
          :math:`[\mathrm{dimensionless}]`
```

mphoward (commented):
This looks great! I have one comment about renaming the `lam` parameter, plus any feedback from @astatt, then we can merge!

src/pair.py
astatt: I agree! `attraction_scale_factor` is good!

astatt (approved):
Ready to merge!

mphoward:
Thanks Mohammadreza!

PR #74, opened 2024-06-03
Update Hertz potential to work with HOOMD 4

clpetix:


clpetix:
@mphoward this should be ready for a review!

src/PairEvaluatorHertz.h
```diff
@@ -2,27 +2,48 @@
...
+#ifdef __HIPCC__
 #define DEVICE __device__
 #else
 #define DEVICE
 #endif
 
+namespace hoomd
+    {
 namespace azplugins
     {
```
mphoward, line 20: I'll note a few minor whitespace things. Do you have pre-commit running but it didn't do anything to these?
```suggestion
```

src/PairEvaluatorHertz.h
```diff
@@ -2,27 +2,48 @@
...
 
 namespace detail
     {
+struct PairParametersHertz : public PairParameters
```
mphoward, line 23:
```suggestion

struct PairParametersHertz : public PairParameters
```

src/PairEvaluatorHertz.h
```diff
@@ -2,27 +2,48 @@
...
+        }
+
+    pybind11::dict asDict()
+        {
+        pybind11::dict v;
+        v["epsilon"] = epsilon;
+        return v;
+        }
+#endif // __HIPCC__
+    Scalar epsilon;
```
mphoward, line 40:
```suggestion

    Scalar epsilon;
```

src/pair.py
```diff
@@ -3,3 +3,25 @@
...
 """Pair potentials."""
+
+from hoomd.azplugins import _azplugins
+from hoomd.data.parameterdicts import TypeParameterDict
+from hoomd.data.typeparam import TypeParameter
+from hoomd.md import pair
+
+
+class Hertz(pair.Pair):
+    """Hertz potential."""
```
mphoward, line 14: Please begin to copy over some class documentation from old code! Check in with me about what to keep vs. remove.

mphoward (changes requested):
I noted a few minor formatting things, and this needs documentation copied but overall looks great!

src/pair.py
```diff
@@ -3,3 +3,66 @@
...
+        \begin{eqnarray*}
+        V(r)  &= \varepsilon ( 1-\frac{ r }{ r_{\rm{cut}} } )^{5/2} ,
+                & r < r_{\rm{cut}} \\
+              &= 0,& r \ge r_{\rm{cut}}
+        \end{eqnarray*}
+
+    Example::
+
+        nl = hoomd.md.nlist.cell()
+        hertz = azplugins.pair.hertz(r_cut=3.0, nlist=nl)
```
astatt, line 35: The hoomd4 convention appears to be to capitalize potentials now, so if we want to be consistent with hoomd4 naming scheme, it might make sense to cal this potential `Hertzian` instead of `hertz`. See: https://hoomd-blue.readthedocs.io/en/v4.7.0/module-md-pair.html - This change would only affects the python naming scheme.

src/PairEvaluatorHertz.h
```diff
@@ -2,27 +2,49 @@
...
-
 namespace detail
     {
+
+struct PairParametersHertz : public PairParameters
+    {
+#ifndef __HIPCC__
+    PairParametersHertz() : epsilon(0) { }
+
+    PairParametersHertz(pybind11::dict v, bool managed = false)
```
astatt, line 28: Calling this `v` is very non-descriptive, but since it is so localized, I don't mind/care. Are there other pair potentials where we could look to make the name of the dictionary consistent? Or is `v` the consistent name we are using throughout?- if so, fine

astatt (changes requested):
Excellent! I left two small comments regarding naming conventions, but otherwise good to go!

src/pair.py
clpetix: The lowercase "hertz" was a typo. I've corrected those to "Hertz" to be consistent with hoomd4's naming scheme. I think we should be okay keeping Hertz rather than Hertzian.

src/PairEvaluatorHertz.h
clpetix: `v` is used by HOOMD and by @mphoward in the base class

src/pytest/test_pair.py
```diff
@@ -17,6 +18,46 @@
 
 potential_tests = []
 
+# pair Hertz
```
mphoward, line 21:
```suggestion
# Hertz
```

src/pytest/test_pair.py
```diff
@@ -17,6 +18,46 @@
...
+potential_tests += [
+    PotentialTestCase(
```
mphoward, line 23: Can we mix in short comment lines for each of these about why they were chosen, for future reference?

mphoward (approved):

src/pair.py
astatt: Yep, sounds good!

src/PairEvaluatorHertz.h
astatt: Ok, great, then we will use `v` consistently everywhere.

astatt (approved):

src/pytest/test_pair.py
astatt: I think that's a great idea - it will help future development to see why/what is tested.

clpetix:
This should be ready to merge, @mphoward!

mphoward:
Thanks Levi!

PR #75, opened 2024-06-03
Refactor double well bond potential

mphoward:
@clpetix Can you polish this one off with documentation when you get a chance?


clpetix:
On it!

src/BondEvaluatorDoubleWell.h
```diff
@@ -2,108 +2,95 @@
...
         }
+#endif
 
-    //! Accept the optional diameter values
-    /*!
-     * \param da Diameter of particle a
-     * \param db Diameter of particle b
-     */
-    DEVICE void setDiameter(Scalar da, Scalar db) { }
+    Scalar V_max; //!< Potential difference between the the first minima and maxima
```
astatt, line 52: Naming suggestion: 
`V_max` -> `V_barrier` = Potential barrier height between the two minima as measured between the first minimum and maximum 
`a/2` -> `r_tr` location of the transition state where the potential barrier is maximum
`b` ->  `r_w` width of the double well potential 
`c` -> `V_a`  Potential height in the activated state with respect to the first minimum 

Honestly, it might make sense to redefine the parameters all-together. `V_max, a,b,c` are the "native" variables going into the potential, but it might be more user friendly to define the location of the first minimum (`a/2-b`), and `a/2` (location of the transition state) instead of `a` and `b`. Then we can calculate the necessary parameters from the user input on the python level. But it is also ok as is, a matter of taste/preference...

astatt: `V_a` could also be `V_tilt` or `V_offset`...

mphoward: I am good with whatever you like here, since this is mostly your potential! If you can make a recommendation, we can implement it.

astatt:
@clpetix  I apologize for creating more work for you! After talking to my student (the user of this potential), I think it makes the most sense to define:

-  location of the first minimum `(a/2-b)`
- `a/2` (location of the transition state)

 instead of `a` and `b`, i.e redefining the variables a bit. 

Then I suggest the following naming/documentation: 

- `V_barrier`  Potential barrier height between the two minima as measured between the first minimum and maximum
-  `r_tr` location of the transition state where the potential barrier is maximum
- `r_0`  location of the first minimum
- `V_offset` Potential height in the activated state with respect to the first minimum

Then, in the documentation we can mention the equation for the location of the second minimum etc. 

Does this sound good and makes sense? Feel free to ping me if you need input/help!

src/BondEvaluatorDoubleWell.h
mphoward: I thought about this and:
* I would rename the variables based on the minimum and transition state _only if_ we remove the `c` parameter. I think `c` makes these values change, right?
* If we don't remove the `c` parameter, I think we should just leave things more or less as is, and we can add some documentation suggesting how to choose them
@astatt let me know which you prefer so we can finalize this today.

clpetix:
We updated the unit tests to make sense. The new implementation means that the differences of `r`, `r0`, and `r1` could all be shifted arbitrarily without messing with the energies or forces. 

This should be ready for y'all to review @astatt @mphoward !

src/BondEvaluatorDoubleWell.h
```diff
@@ -2,151 +2,132 @@
...
+        pybind11::dict v;
+        v["r_0"] = r_1 - r_diff;
+        v["r_1"] = r_1;
+        v["U_1"] = U_1;
+        v["U_tilt"] = U_tilt;
+        return v;
         }
+#endif
+
+    Scalar r_1; //!<
```
astatt, line 52: add documentation describing briefly what the variables mean

src/BondEvaluatorDoubleWell.h
```diff
@@ -2,151 +2,132 @@
...
+    __attribute__((aligned(32)));
+#endif
 
-    //! Accept the optional diameter values
-    /*!
-     * \param da Diameter of particle a
-     * \param db Diameter of particle b
-     */
-    DEVICE void setDiameter(Scalar da, Scalar db) { }
+//! Class for evaluating the double well bond potential
```
astatt, line 63: This documentation needs to be updated  - it can be left for later, but then we need to leave a TODO note at least

src/bond.py
```diff
@@ -3,3 +3,65 @@
...
+    `DoubleWell` specifies a double well potential between the two particles in
+    the simulation state with:
+
+    .. math::
+
+        U(r)  =  \frac{V_{max}-c/2}{b^4} \left[
+            \left( r - a/2 \right)^2 - b^2 \right]^2
+            +\frac{c}{2b}(r - a/2) + c/2
+
+    Attributes:
```
astatt, line 29: Documentation needs to be updated. Same comment as above, can be left for later but we should note that somewhere with a TODO.

astatt (changes requested):

src/bond.py
```diff
@@ -3,3 +3,66 @@
...
+            * ``r_1`` (`float`, **required**) - The location of the
+                local potential maximum( in distance units)
+
+            * ``U_1`` (`float`, **required**) - Potential maximum energy
+                barrier between the two minima at ``r_1`` for ``U_tilt=0`` (in
+                energy units)
+
+            * ``U_tilt`` (`float`, **required**) -tunes the energy offset
+                between the two potential minima values, i.e. it tilts the
+                potential (in energy units). The default value of c is zero.
```
astatt, line 48: default value of c -> default value of U_tilt

astatt (approved):
Found one typo in the python documentation :) Otherwise looks great!

src/BondEvaluatorDoubleWell.h
```diff
@@ -2,151 +2,133 @@
...
-        bond_eng = ((V_max - c_half) / (b_sq * b_sq)) * d * d + c_half / b * r_min_half_a + c_half;
-        force_divr = -(4 * (V_max - c_half) / (b_sq * b_sq) * d * r_min_half_a + c_half / b) / r;
+        const Scalar r = fast::sqrt(rsq);
+        const Scalar x = (r_1 - r) / r_diff;
+        const Scalar x2 = x * x;
+        const Scalar y = Scalar(1.0) - x2;
+        const Scalar y2 = y * y;
 
+        bond_eng = (U_1 * y2 + U_tilt * (Scalar(1.0) - x - y2));
+        force_divr = ((Scalar(4.0) * x * y * (U_tilt - U_1) - U_tilt) / r_diff) / r;
```
mphoward, line 111:
```suggestion
        bond_eng = U_1 * y2 + U_tilt * (Scalar(1.0) - x - y2);
        force_divr = (Scalar(4.0) * x * y * (U_tilt - U_1) - U_tilt) / (r_diff * r);
```

src/bond.py
```diff
@@ -3,3 +3,66 @@
...
+            * ``r_1`` (`float`, **required**) - The location of the
+                local potential maximum( in distance units)
+
+            * ``U_1`` (`float`, **required**) - Potential maximum energy
+                barrier between the two minima at ``r_1`` for ``U_tilt=0`` (in
+                energy units)
+
+            * ``U_tilt`` (`float`, **required**) -tunes the energy offset
+                between the two potential minima values, i.e. it tilts the
+                potential (in energy units). The default value of U_tilt is zero.
```
mphoward, line 48: This needs some updating to match HOOMD convention, but I will commit on it directly since it's hard to type in the web interface.

src/pytest/test_bond.py
```diff
@@ -3,3 +3,128 @@
...
+        hoomd.azplugins.bond.DoubleWell,
+        dict(r_0=1.0, r_1=2.0, U_1=1.0, U_tilt=0.5),
+        2.5,
+        1.03125,
+        0.25,
+    ),
+]
+
+
+@pytest.fixture(scope='session')
```
mphoward, line 72: I am going to try to move this to a `contest.py` since it might be helpful to have available outside this file.

mphoward (commented):
Looks great! I am going to check on the things below then merge.

PR #76, opened 2024-06-03
Refactor two-patch Morse potential

mphoward:


src/AnisoPairEvaluatorTwoPatchMorse.h
```diff
@@ -2,37 +2,59 @@
...
 
-/*!
- * \file AnisoPairEvaluatorTwoPatchMorse.h
- * \brief Defines the aniostropic pair force evaluator class for Two-patch Morse potential
- */
-
 #ifndef AZPLUGINS_ANISO_PAIR_EVALUATOR_TWO_PATCH_MORSE_H_
 #define AZPLUGINS_ANISO_PAIR_EVALUATOR_TWO_PATCH_MORSE_H_
 
 #include "AnisoPairEvaluator.h"
```
mphoward, line 8: The `<string>` header should be included since we use it. It is working now because it comes from somewhere else, but it's a good idea to "include what you use" if possible.
```suggestion
#ifndef __HIPCC__
#include <string>
#endif // __HIPCC__

#include "AnisoPairEvaluator.h"
```

src/AnisoPairEvaluatorTwoPatchMorse.h
```diff
@@ -2,37 +2,59 @@
...
 #define HOSTDEVICE
 #endif
 
+namespace hoomd
+    {
 namespace azplugins
     {
-
 namespace detail
     {
```
mphoward, line 25:
```suggestion
    {

```

src/AnisoPairEvaluatorTwoPatchMorse.h
```diff
@@ -2,37 +2,59 @@
...
 //! Two-patch Morse parameters
-/*!
- * \sa AnisoPairEvaluatorTwoPatchMorse
- */
-struct two_patch_morse_params : public AnisoPairParams
+struct AnisoPairParametersTwoPatchMorse : public AnisoPairParameters
     {
+#ifndef __HIPCC__
+    AnisoPairParametersTwoPatchMorse()
+        : Mdeps(0), Mrinv(0), req(0), omega(0), alpha(0), repulsion(false)
```
mphoward, line 31: Since `Mrinv` is an inverse quantity, it can't be zero... let's default it 1.0.
```suggestion
        : Mdeps(0), Mrinv(1.0), req(0), omega(0), alpha(0), repulsion(false)
```

src/pair.py
```diff
@@ -143,3 +143,96 @@ def __init__(self, nlist, default_r_cut=None, default_r_on=0, mode='none'):
             ),
         )
         self._add_typeparam(params)
+
+
+class TwoPatchMorse(pair.aniso.AnisotropicPair):
+    r"""TwoPatchMorse potential.
```
mphoward, line 149:
```suggestion
    r"""Two-patch Morse potential.
```

src/pair.py
```diff
@@ -143,3 +143,96 @@ def __init__(self, nlist, default_r_cut=None, default_r_on=0, mode='none'):
...
+
+    Args:
+        nlist (hoomd.md.nlist.NeighborList): Neighbor list.
+        default_r_cut (float): Default cutoff radius :math:`[\mathrm{length}]`.
+        mode (str): Energy shifting/smoothing mode.
```
mphoward, line 154:
```suggestion
        mode (str): Energy shifting mode.
```

src/pair.py
```diff
@@ -143,3 +143,96 @@ def __init__(self, nlist, default_r_cut=None, default_r_on=0, mode='none'):
...
+        The Two Patch Morse potential parameters. The dictionary has the following
+        keys:
+
+        * ``Mdeps`` (`float`, **required**) - controls the depth of the
+         potential well
+          :math:`M_d` :math:`[\mathrm{energy}]`
+        * ``Mr`` (`float`, **required**) - controls the width of the
+         potential well :math:`M_r` :math:`[\mathrm{length}]`
+        * ``req`` (`float`, **required**) - controls the position of
+        the potential well :math:`r_eq` :math:`[\mathrm{length}]`
```
mphoward, line 206: I think we should rename these parameters based on how the equations are structured (not sure why they were the way they were before). Please update here, and then also in the header file.
```suggestion
        * ``M_d`` (`float`, **required**) - controls the depth of the
         potential well
          :math:`M_d` :math:`[\mathrm{energy}]`
        * ``M_r`` (`float`, **required**) - controls the width of the
         potential well :math:`M_r` :math:`[\mathrm{length}]`
        * ``r_eq`` (`float`, **required**) - controls the position of
        the potential well :math:`r_eq` :math:`[\mathrm{length}]`
```

src/pair.py
```diff
@@ -143,3 +143,96 @@ def __init__(self, nlist, default_r_cut=None, default_r_on=0, mode='none'):
...
+              -\frac{r-r_{\rm eq}}{M_r}\right) \right]^2 - 1 \right)
+
+    and :math:`\Omega(\gamma)` depends on the orientations
+
+    .. math::
+        \Omega(\gamma) = \frac{1}{1+\exp[-\omega (\gamma^2 - \alpha)]}
+
+    The potential can be smoothed to zero force (making it purely attractive)
+    when :math:`r < r_{\rm eq}` by making :math:`V_{\rm M}(r < r_{\rm eq}) = -M_d`
+    when the option  *repulsion* is ``False``.
```
mphoward, line 178:
```suggestion
    when the option  ``repulsion`` is ``False``.
```

src/AnisoPairEvaluatorTwoPatchMorse.h
```diff
@@ -120,17 +113,12 @@ class AnisoPairEvaluatorTwoPatchMorse : public AnisoPairEvaluator
...
-     * cutoff.
-     *
-     */
+    //! Whether the potential implements the energy_shift parameter
+    HOSTDEVICE static bool constexpr implementsEnergyShift()
+        {
+        return true;
+        }
+
     DEVICE bool evaluate(Scalar3& force,
```
mphoward, line 122: I am leaving a note to myself to review this method before merging. You don't need to touch it, beyond updating the parameter names as noted below.

src/pytest/test_pair_aniso.py
```diff
@@ -0,0 +1,150 @@
...
+    # test without potential shifting
+    PotentialTestCase(
+        hoomd.azplugins.pair.TwoPatchMorse,
+        {
+            'Mdeps': 1.8341,
+            'Mr': 0.0302,
+            'req': 1.0043,
+            'omega': 5.0,
+            'alpha': 0.40,
+            'repulsion': False,
```
mphoward, line 32: Were there any test cases in the original file where `repulsion is True`?

mphoward (commented):
Thanks Mohammadreza! Can you clean up the items noted below, and let me know when you're done with that so I can have a look at the thing I noted?

src/pytest/test_pair_aniso.py
mzf0069: There is one. But the positions are different.

src/pair.py
mzf0069: How would you like Mrinv to be renamed? M_rinv?

mphoward:
Thanks Mohammadreza!

PR #77, opened 2024-06-03
Update Colloid potential to work with HOOMD4. 

clpetix:


clpetix:
@mphoward @astatt this code is ready for y'all to take a look at! The documentation will need cleaned up, but the actual potential is working and recreates the old tests!

src/PairEvaluatorColloid.h
```diff
@@ -255,42 +269,40 @@ class PairEvaluatorColloid : public PairEvaluator
         // compute the force divided by r in force_divr
         if (rsq < rcutsq && A != 0)
             {
-            if (form == SOLVENT_SOLVENT)
+            // if (form == SOLVENT_SOLVENT)
```
astatt, line 272: I would change this comment to simply read: // Solvent-Solvent interaction

src/PairEvaluatorColloid.h
```diff
@@ -255,42 +269,40 @@ class PairEvaluatorColloid : public PairEvaluator
...
-            else if (form == COLLOID_SOLVENT)
-                {
-                pair_eng = computeColloidSolvent<true>(force_divr, rsq);
-                if (energy_shift)
-                    {
-                    pair_eng -= computeColloidSolvent<false>(force_divr, rcutsq);
-                    }
-                }
-            else if (form == COLLOID_COLLOID)
+            // else if (form == COLLOID_COLLOID)
```
astatt, line 281: same as above

src/PairEvaluatorColloid.h
```diff
@@ -255,42 +269,40 @@ class PairEvaluatorColloid : public PairEvaluator
...
+            else if (ai != 0 && aj != 0)
                 {
                 pair_eng = computeColloidColloid<true>(force_divr, rsq);
                 if (energy_shift)
                     {
                     pair_eng -= computeColloidColloid<false>(force_divr, rcutsq);
                     }
                 }
+            // else if (form == COLLOID_SOLVENT)
```
astatt, line 290: and again :)

src/PairEvaluatorColloid.h
```diff
@@ -2,27 +2,58 @@
...
+    {
 namespace azplugins
     {
 namespace detail
     {
 
+struct PairParametersColloid : public PairParameters
+    {
+#ifndef __HIPCC__
+    PairParametersColloid() : A(0), a1(0), a2(0), sigma_3(0) { }
```
astatt, line 26: Is there a reason for `sigma_3` vs `sigma` in the names of the parameters?

src/pytest/test_pair.py
```diff
@@ -150,6 +150,40 @@
     ),
 ]
 
+# Colloid
+potential_tests += [
```
astatt, line 154: Add a test case outside of the cutoff where Force and Energy should be zero. Just for completeness sake.

astatt (changes requested):

astatt:
Looks good! I have some minor comments, then it is good to go!

src/PairEvaluatorColloid.h
clpetix: `sigma_3` (sigma^3) is used to calculate the energies and forces, whereas `sigma` is never used directly. I cleaned up the implementation a bit to make it clearer!

src/PairEvaluatorColloid.h
```diff
@@ -2,27 +2,59 @@
...
+    Scalar A;
+    Scalar a_1;
+    Scalar a_2;
+    Scalar sigma_3;
+    }
+#if HOOMD_LONGREAL_SIZE == 32
+    __attribute__((aligned(16)));
+#else
+    __attribute__((aligned(32)));
+#endif
```
mphoward, line 57:
```suggestion
#endif

```

src/PairEvaluatorColloid.h
```diff
@@ -2,27 +2,59 @@
...
+        v["a_2"] = a_2;
+        v["sigma"] = std::cbrt(sigma_3);
+        return v;
+        }
+#endif // __HIPCC__
+
+    Scalar A;
+    Scalar a_1;
+    Scalar a_2;
+    Scalar sigma_3;
```
mphoward, line 51:
```suggestion
    Scalar A; //!< Hamaker constant
    Scalar a_1; //!< particle 1 radius
    Scalar a_2; //!< particle 2 radius
    Scalar sigma_3; //!< Lennard-Jones sigma
```

src/PairEvaluatorColloid.h
```diff
@@ -34,44 +66,35 @@ namespace detail
...
  *
  * The strength of all potentials is set by the Hamaker constant, represented here by the
  * symbol \f$A\f$. The other parameter \f$\sigma\f$ is the diameter of the particles that
  * are integrated out (colloids are comprised of Lennard-Jones particles with parameter
  * \f$\sigma\f$). The parameters that are fed in are:
  *
  * - \a A - the Hamaker constant
+ * - \a a_1 - diameter of particle 1
+ * - \a a_2 - diameter of particle 2
  * - \a sigma_3 - \f$\sigma^3\f$
```
mphoward, line 91: I think this discussion can be removed. Some of it isn't quite right anymore, and I would prefer just to keep the Python docs pristine.

src/PairEvaluatorColloid.h
```diff
@@ -34,44 +66,35 @@ namespace detail
...
- * - \a sigma_6 - \f$\sigma^6\f$
- * - \a form - the style of the pair interaction as in int that is cast to the interaction_type
  */
 class PairEvaluatorColloid : public PairEvaluator
     {
     public:
     //! Define the parameter type used by this pair potential evaluator
```
mphoward, line 96:
```suggestion
```

src/PairEvaluatorColloid.h
```diff
@@ -255,42 +261,40 @@ class PairEvaluatorColloid : public PairEvaluator
         // compute the force divided by r in force_divr
         if (rsq < rcutsq && A != 0)
             {
-            if (form == SOLVENT_SOLVENT)
+            // Solvent-Solvent interaction
+            if (ai == 0 && aj == 0)
```
mphoward, line 265: Wrap all these zeros here & below in `Scalar` to make sure there's no type casting.
```suggestion
            if (ai == Scalar(0) && aj == Scalar(0))
```

src/PairEvaluatorColloid.h
```diff
@@ -302,17 +306,17 @@ class PairEvaluatorColloid : public PairEvaluator
 #endif
 
     protected:
-    Scalar A;              //!< Hamaker constant
-    Scalar sigma_3;        //!< Sigma^3
-    Scalar sigma_6;        //!< Sigma^6
-    interaction_type form; //!< Form of the interaction
+    Scalar A;       //!< Hamaker constant
+    Scalar sigma_3; //!< Sigma^3
+    Scalar sigma_6; //!< Sigma^6
```
mphoward, line 312:
```suggestion
```

src/PairEvaluatorColloid.h
astatt: thank you!

src/pair.py
```diff
@@ -10,6 +10,79 @@
...
+    sphere and a sphere. The attractive part of the colloid-colloid pair
+    potential was derived originally by Hamaker, and the full potential by
+    `Everaers and Ejtehadi <http://doi.org/10.1103/PhysRevE.67.041710>`_.
+    A discussion of the application of these potentials to colloidal suspensions
+    can be found in `Grest et al. <http://dx.doi.org/10.1063/1.3578181>`_
+
+    The pair potential has three different coupling styles between particle types:
+
+    - Solvent-Solvent gives the Lennard-Jones potential for coupling between
+    pointlike particles
```
mphoward, line 32:
```suggestion
    - When the radius of both particles is zero, the potential is the usual Lennard-Jones coupling:
```

src/pair.py
```diff
@@ -10,6 +10,79 @@
...
+                  - \left(\frac{\sigma}{r}\right)^6 \right] & r < r_{\mathrm{cut}} \\
+              = & 0 & r \ge r_{\mathrm{cut}}
+        \end{eqnarray*}
+
+    - Colloid-Solvent gives the interaction between a pointlike particle and a colloid
+    - Colloid-Colloid gives the interaction between two colloids
+
+    Refer to the work by `Grest et al. <http://dx.doi.org/10.1063/1.3578181>`_ for the
+    form of the colloid-solvent and colloid-colloid potentials, which are too cumbersome
+    to report on here.
```
mphoward, line 48: I think we can make these comments a little more helpful in two ways:
1. Describe the cases using the radii (rather than colloid-solvent and colloid-colloid)
2. Can we cite equation numbers from the Grest paper that the reader can refer to? I think you can stick the url in one place and refer to it multiple times, like here: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#links-to-external-web-pages .


It's Eqs. (3) & (4) for the colloid-colloid potential, and Eq. (5) for the colloid-solvent.

src/pair.py
```diff
@@ -10,6 +10,79 @@
...
+
+    .. warning::
+        The diameter parameters are used to infer which case is used to compute
+        the interactions. A particle diameter equal to 0 is used to infer
+        a solvent interaction. You must make sure you appropriately set the
+        diameter.
```
mphoward, line 54: This warning was a leftover from the old code. I think we don't need it since the user sets the radii explicitly, and we are documenting above what happens in these cases.
```suggestion
```

src/pair.py
```diff
@@ -10,6 +10,79 @@
...
+
```
mphoward, line 55: Should we also add some comments/equations about how the Hamaker constant should be chosen in terms of the LJ interactions? The Grest paper also gives the formulas for those in terms of the LJ epsilon and sigma, plus the nominal LJ density of the spheres.

src/pair.py
```diff
@@ -10,6 +10,79 @@
...
+    Example::
```
mphoward, line 56: I think we should add a short sentence explaining what system this example is setting up, and possibly refer to my paper.

mphoward (changes requested):
Thanks Levi! This looks great, and I have mostly comments on the documentation. Please make sure to pull before making these changes. I resolved merge conflicts and tried a change to get the MPI tests to pass (box was too small).

src/pair.py
mphoward: Sounds good!

clpetix:
@mphoward, I got all the changes implemented. This is ready for you again!

mphoward (approved):
Looks great, thank you!

PR #79, opened 2024-06-04
Refactor integrator

mayukh33:


src/TwoStepBrownianFlow.h
```diff
@@ -90,10 +84,6 @@ template<class FlowField> class PYBIND11_EXPORT TwoStepBrownianFlow : public Two
     /*!
      * \param noiseless If true, do not apply a random diffusive force
      */
-    void setNoiseless(bool noiseless)
```
mphoward, line 93: This setter should not have been removed, please restore.

src/TwoStepBrownianFlow.h
```diff
@@ -192,19 +167,16 @@ void export_TwoStepBrownianFlow(pybind11::module& m, const std::string& name)
...
                       std::shared_ptr<Variant>,
                       std::shared_ptr<FlowField>,
-                      unsigned int,
-                      bool,
-                      Scalar,
                       bool>())
-        .def("setFlowField", &BrownianFlow::setFlowField)
-        .def("setNoiseless", &BrownianFlow::setNoiseless);
+        .def("getFlowField", &BrownianFlow::getFlowField)
+        .def("getNoiseless", &BrownianFlow::getNoiseless);
```
mphoward, line 177: This is how you make a property. `flow_field` should be read-only, but `noiseless` can be mutable.
```suggestion
        .def_property_readonly("flow_field", &BrownianFlow::getFlowField)
        .def_property("noiseless", &BrownianFlow::getNoiseless, &BrownianFlow::setNoiseless);
```

src/TwoStepBrownianFlow.h
```diff
@@ -192,19 +167,16 @@ void export_TwoStepBrownianFlow(pybind11::module& m, const std::string& name)
...
     }
     } // end namespace detail
     } // end namespace azplugins
-
+    } // end namespace hoomd
```
mphoward, line 181:
```suggestion
    } // end namespace hoomd

```

src/TwoStepLangevinFlow.h
mphoward, line 1: Please apply the same changes to this file as for the BD one.

src/module.cc
mphoward, line 1: I will help you write some machinery to compile these.

mphoward (changes requested):
Thanks, this is going well too! Can you please take care of the small comments here, and then I can help you with the CMake setup for compiling these so you can also work on the GPU code.

PR #81, opened 2024-07-09
Refactor general weight DPD potential

astatt:


astatt:
pre-commit.ci autofix

astatt:
pre-commit.ci autofix

src/pair.py
```diff
@@ -5,9 +5,127 @@
...
+        \end{eqnarray*}
+
+    The conservative force :math:`\mathbf{F}_{\rm C}` is the standard form:
+
+    .. math::
+        :nowrap:
+        \begin{eqnarray*}
+        \mathbf{F}_{\rm C} =& A (1- r_{ij}/r_{\rm cut}) & r \le r_{\rm cut} \\
+                           =& 0 & r > r_{\rm cut}
+        \end{eqnarray*}
```
mzf0069, line 41: I think `r` should be changed to `r_{ij}`:
```
.. math::
        :nowrap:
        \begin{eqnarray*}
        \mathbf{F}_{\rm C} =& A (1- r_{ij}/r_{\rm cut}) & r_{ij} \le r_{\rm cut} \\
                           =& 0 & r_{ij} > r_{\rm cut}
        \end{eqnarray*}
```

src/pair.py
```diff
@@ -5,9 +5,127 @@
...
+          (\mathbf{v}_{ij} \cdot \mathbf{\hat r}_{ij}) \mathbf{\hat r}_{ij} \\
+        \mathbf{F}_{\rm R} =& \sigma \omega_{\rm R}(r_{ij}) \xi_{ij}
+            \mathbf{\hat r}_{ij}
+        \end{eqnarray*}
+
+    where :math:`\sigma = 2\gamma k_{\rm B}T` and
+    :math:`\omega_{\rm D} = \left[\omega_{\rm R} \right]^2`
+    to satisfy the fluctuation dissipation relation. The genealized weight
+    function is given by the form proposed by
+    `Fan et al. <https://doi.org/10.1063/1.2206595>`_:
```
mzf0069, line 63: I think this needs to be expanded based on  `Fan et al. <https://doi.org/10.1063/1.2206595>`. Also, I think the correct relation between sigma and gamma is :math:`\sigma ^2 = 2\gamma k_{\rm B}T` :
```
where :math:`\sigma` and :math:`gamma` are two coefficients characterizing the strengths of the dissipative and random forces. To satisfy the fluctuation dissipation relation, :math:`\omega_{\rm D} = \left[\omega_{\rm R} \right]^2` where  :math:`\sigma ^2 = 2\gamma k_{\rm B}T`. Here, :math:`v_{ij} = v_i - v_j`, and :math:`\xi_{ij}` is a white noise function. The generalized weight function is given by the form proposed by
`Fan et al. <https://doi.org/10.1063/1.2206595>`_: 
```

src/pair.py
```diff
@@ -5,9 +5,127 @@
...
+
+    .. math::
+        :nowrap:
+        \begin{eqnarray*}
+        w_{\rm D}(r) = &\left( 1 - r/r_{\mathrm{cut}} \right)^s
+                       & r \le r_{\mathrm{cut}} \\
+                     = & 0 & r > r_{\mathrm{cut}} \\
+        \end{eqnarray*}
```
mzf0069, line 71: To be consistent: 
```
.. math::
        :nowrap:
        \begin{eqnarray*}
        w_{\rm D}(r_{ij}) = &\left( 1 - r_{ij}/r_{\mathrm{cut}} \right)^s
                       & r_{ij} \le r_{\mathrm{cut}} \\
                     = & 0 & r_{ij} > r_{\mathrm{cut}} \\
        \end{eqnarray*}
```

mphoward:
I noticed that `DPDGeneralWeight` needs to be added to the sphinx doc pages too. It isn't rendering on RTD currently.

clpetix (commented):
The bulk of the code looks good to me! 

Like Mike mentioned, the `DPDGeneralWeight` needs added to the documentation here: 

https://github.com/mphowardlab/azplugins/blob/d5fc5cf7df2fd6cfbf6b5ab2e949f0c58f5046a5/doc/module-azplugins-pair.rst?plain=1#L18

and then check that once it builds. 

What is our reasoning behind having a separate `test_pair_dpd.py` file with tests being split between it and `test_pair.py`? Would it be clearer to have all these tests consolidated into one file like we did for aniso?

mphoward:
> What is our reasoning behind having a separate `test_pair_dpd.py` file with tests being split between it and `test_pair.py`? Would it be clearer to have all these tests consolidated into one file like we did for aniso?

No particular reason other than it was that way before, and it may be slightly more readable separated out. The aniso pair tests are in a separate file for a similar reason. But, it can be moved if you think that’s better!

mzf0069:
The code looks good to me. I made a couple of comments on the documentation above.

clpetix:
> > What is our reasoning behind having a separate `test_pair_dpd.py` file with tests being split between it and `test_pair.py`? Would it be clearer to have all these tests consolidated into one file like we did for aniso?
> 
> No particular reason other than it was that way before, and it may be slightly more readable separated out. The aniso pair tests are in a separate file for a similar reason. But, it can be moved if you think that’s better!

Gotcha! I see pros and cons to both options, but ultimately, I don't have a strong preference. I'm good with whatever you all think is best!

mphoward:
Lacking a preference either way, let's leave as is then.

mphoward:
Hi all, the documentation should be cleaned up on this, so please give it a final review before merging!

PR #82, opened 2024-07-18
Quartic bond potential

bcjeong1616:
Created the quartic bond potential and updated to be compatible with Hoomd 4.x installations.

Tested build details:
Python v3.12.3
Hoomd v4.6.0
gcc v9
pybind11 v2.10.4
eigen v3.4.0

Created tests for the potential
Passed all tests

### DIFF
--- a/src/BondEvaluatorQuartic.h
+++ b/src/BondEvaluatorQuartic.h
@@ -113,12 +113,17 @@ class BondEvaluatorQuartic : public BondEvaluator
     public:
     typedef BondParametersQuartic param_type;
 
-    DEVICE BondEvaluatorQuartic(Scalar _rsq, const param_type& _params)
-        : BondEvaluator(_rsq), k(_params.k), r_0(_params.r_0), b_1(_params.b_1), b_2(_params.b_2),
-          U_0(_params.U_0), delta(_params.delta),
-          lj1(_params.epsilon_x_4 * _params.sigma_6 * _params.sigma_6),
-          lj2(_params.epsilon_x_4 * _params.sigma_6), epsilon(_params.epsilon_x_4 / Scalar(4.0))
+    DEVICE BondEvaluatorQuartic(Scalar _rsq, const param_type& _params) : BondEvaluator(_rsq)
         {
+        k = _params.k;
+        r_0 = _params.r_0;
+        b_1 = _params.b_1;
+        b_2 = _params.b_2;
+        U_0 = _params.U_0;
+        delta = _params.delta;
+        lj1 = _params.epsilon_x_4 * _params.sigma_6 * _params.sigma_6;
+        lj2 = _params.epsilon_x_4 * _params.sigma_6;
+        epsilon = _params.epsilon_x_4 / Scalar(4.0);
         }
 
     DEVICE bool evalForceAndEnergy(Scalar& force_divr, Scalar& bond_eng)
@@ -218,8 +223,8 @@ class BondEvaluatorQuartic : public BondEvaluator
     private:
     Scalar k;       //!< k parameter for quartic potential strength
     Scalar r_0;     //!< energy barrier breaking distance
-    Scalar b_1;     //!< quartic tuning parameter #1
-    Scalar b_2;     //!< quartic tuning parameter #2
+    Scalar b_1;     //!< primary quartic potential tuning parameter
+    Scalar b_2;     //!< secondary quartic potential tuning parameter
     Scalar U_0;     //!< quartic energy barrier to "breaking"
     Scalar delta;   //!< delta parameter bond
     Scalar lj1;     //!< lj1 parameter used in WCA calculation
--- a/src/BondEvaluatorQuartic.h
+++ b/src/BondEvaluatorQuartic.h
@@ -28,7 +28,10 @@ namespace detail
 struct BondParametersQuartic : public BondParameters
     {
 #ifndef __HIPCC__
-    BondParametersQuartic() : k(0), r_0(0), b_1(0), b_2(0), U_0(0), sigma_6(0), epsilon_x_4(0), delta(0) { }
+    BondParametersQuartic()
+        : k(0), r_0(0), b_1(0), b_2(0), U_0(0), sigma_6(0), epsilon_x_4(0), delta(0)
+        {
+        }
 
     BondParametersQuartic(pybind11::dict v)
         {
@@ -69,7 +72,7 @@ struct BondParametersQuartic : public BondParameters
     Scalar b_2;         //!< quartic tuning parameter #2
     Scalar U_0;         //!< quartic energy barrier to "breaking"
     Scalar sigma_6;     //!< sigma raised to the power of 6 for WCA
-    Scalar epsilon_x_4; //!< epsilon * 4 for WCA 
+    Scalar epsilon_x_4; //!< epsilon * 4 for WCA
     Scalar delta;       //!< delta parameter bond
     }
 #if HOOMD_LONGREAL_SIZE == 32
@@ -81,13 +84,13 @@ struct BondParametersQuartic : public BondParameters
 //! Class for evaluating the quartic bond potential
 /*!
  * This bond potential follows the functional form
- * 
+ *
  * \begin{eqnarray*}
- *      U(r) = & k4 ((r - r_{0}) - b_{1})((r - r_{0}) - b_{2})(r - r_{0})^2 
+ *      U(r) = & k4 ((r - r_{0}) - b_{1})((r - r_{0}) - b_{2})(r - r_{0})^2
  *                 + U_{0} + U_{\rm WCA}(r)      & r < r_0\\
  *            = & U_0                            & r \ge r_0
  * \end{eqnarray*}
- * 
+ *
  * which has a WCA repulsive force at close range, and a quartic energy well. On the far side of
  * this energy well, there is an energy barrier, to a plateau energy, U_0. It reaches this energy
  * at r_0. Beyond r_0, the energy is constant, and there is no force from the bond potential.
@@ -95,7 +98,7 @@ struct BondParametersQuartic : public BondParameters
  * Since this potential has a finite range, beyond which no force is applied, this potential is
  * useful for modeling bonds with scissile functionality. r_0 defines the bond length at which
  * the bond can be considered broken, and U_0 defines the height of the energy barrier to breaking.
- * 
+ *
  * The parameters b_1 and b_2 tune the location of the energy minimum, as well as the well width.
  * They do not cleanly translate to meaningful locations on the potential energy surface, but serve
  * as helpful tuning parameters to mimic other bond topologies. For example, making a previously
@@ -111,11 +114,10 @@ class BondEvaluatorQuartic : public BondEvaluator
     typedef BondParametersQuartic param_type;
 
     DEVICE BondEvaluatorQuartic(Scalar _rsq, const param_type& _params)
-        : BondEvaluator(_rsq), k(_params.k), r_0(_params.r_0), b_1(_params.b_1),
-          b_2(_params.b_2), U_0(_params.U_0), delta(_params.delta), 
+        : BondEvaluator(_rsq), k(_params.k), r_0(_params.r_0), b_1(_params.b_1), b_2(_params.b_2),
+          U_0(_params.U_0), delta(_params.delta),
           lj1(_params.epsilon_x_4 * _params.sigma_6 * _params.sigma_6),
-          lj2(_params.epsilon_x_4 * _params.sigma_6),
-          epsilon(_params.epsilon_x_4 / Scalar(4.0))
+          lj2(_params.epsilon_x_4 * _params.sigma_6), epsilon(_params.epsilon_x_4 / Scalar(4.0))
         {
         }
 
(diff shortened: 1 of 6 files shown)


mphoward:
pre-commit.ci autofix

mphoward:
Please add your new potential to the Sphinx documentation & ensure it compiles without warnings or errors, then run pre-commit locally & fix any issues it identifies before review. Thank you!

src/BondEvaluatorQuartic.h
```diff
@@ -0,0 +1,236 @@
...
+ *
+ * The delta parameter shifts the whole potential with respect to r. The default value is 0.
+ * Positive values shift to higher r values.
+ */
+class BondEvaluatorQuartic : public BondEvaluator
+    {
+    public:
+    typedef BondParametersQuartic param_type;
+
+    DEVICE BondEvaluatorQuartic(Scalar _rsq, const param_type& _params)
```
astatt, line 116: This is a lot of parameters. While equivalent, I would prefer if you could assign them inside the constructor, like for example here: https://github.com/mphowardlab/azplugins/blob/0c520e05afe59ef07ab3405fccaff4d90524fc45/src/DPDPairEvaluatorGeneralWeight.h#L109

src/BondEvaluatorQuartic.h
```diff
@@ -0,0 +1,236 @@
...
+        {
+        return std::string("Quartic");
+        }
+#endif
+
+    private:
+    Scalar k;       //!< k parameter for quartic potential strength
+    Scalar r_0;     //!< energy barrier breaking distance
+    Scalar b_1;     //!< quartic tuning parameter #1
+    Scalar b_2;     //!< quartic tuning parameter #2
```
astatt, line 222: Can these parameters (b_1 and b_2) be described in a way that makes clear which one is which?

src/bond.py
```diff
@@ -63,3 +63,102 @@ def __init__(self):
...
+
+            * ``Delta`` (`float`, **optional**) - bond length shift ``Delta``.
+              Effectively replaces all instances of :math:`r \text{ with } r - \Delta`.
+              The default value is zero.
+              :math:`[\mathrm{length}]`
+
+    Examples::
+
+        quartic = azplugins.bond.Quartic()
+        quartic.params['A-A'] = dict(k=1434.3, r_0=1.5, b_1=-0.7589, b_2=0.0,
```
astatt, line 138: delete one - only one example set of parameters needed.

astatt (changes requested):

src/BondEvaluatorQuartic.h
bcjeong1616: Since the b_1 and b_2 are algebraically equivalent, (if you were to exchange their values, you would get the same potential) I'm not sure there's much merit to distinguishing the two. I've changed the description to "primary" and "secondary" so it's a bit easier to see when scanning the document

src/BondEvaluatorQuartic.h
```diff
@@ -0,0 +1,241 @@
...
+#endif
+
+    Scalar k;           //!< k parameter for quartic potential strength
+    Scalar r_0;         //!< energy barrier breaking distance
+    Scalar b_1;         //!< quartic tuning parameter #1
+    Scalar b_2;         //!< quartic tuning parameter #2
+    Scalar U_0;         //!< quartic energy barrier to "breaking"
+    Scalar sigma_6;     //!< sigma raised to the power of 6 for WCA
+    Scalar epsilon_x_4; //!< epsilon * 4 for WCA
+    Scalar delta;       //!< delta parameter bond
```
mphoward, line 76: Do you have a use case for this `delta` parameter? Core shifting is no longer part of the "standard" WCA in HOOMD, so we could remove it if not to simplify the code below. Let me know, as that will inform my other comments!

bcjeong1616: Hi Michael, thank you for having a look! I just double checked, and yes we do have a use case for it. We use it for consistency, as we have some systems with shifted lennard-jones pair potentials alongside these quartic bonds as a sort of secondary pseudo-pair potential.

src/BondEvaluatorQuartic.h
```diff
@@ -0,0 +1,241 @@
...
+    DEVICE bool evalForceAndEnergy(Scalar& force_divr, Scalar& bond_eng)
+        {
+        bond_eng = 0;
+        force_divr = 0;
+        // check for invalid parameters
+        if (r_0 == Scalar(0.0))
+            return false;
+
+        Scalar r = Scalar(-1.0); // sentinel value for whether the square root has already been
+                                 // taken
```
mphoward, line 138: There is alot of duplicated code in this method. Can you restructure it so that isn't the case? Like:
```
force_divr = 0;
bond_eng = 0;

Scalar r_red(1.0);
if (delta == Scalar(0.0))
    {
    r2inv = Scalar(1.0) / rsq;

    // WCA without delta
    bond_eng += ...;
    force_divr += ...;

    if (rsq < r_0 * r_0)
        r_red = fast::sqrt(rsq) - r_0;
    }
else
    {
    r = fast::sqrt(rsq) - delta;

    // WCA with delta
    bond_eng += ...;
    force_divr += ...;

    if (r < r_0)
        r_red = r - r_0;
    }

// bond potential is on when r_red < 0
if (r_red < Scalar(0))
    {
    bond_eng += ...;
    }
else
    {
    bond_eng += U_0;
    }
```
I think the bond potential equations are the same, if you insert delta = 0 in the more general one.

src/bond.py
```diff
@@ -63,3 +63,100 @@ def __init__(self):
...
+    .. math::
+        \begin{eqnarray*}
+        U(r) = & k [(r - r_{0}) - b_{1}][(r - r_{0}) - b_{2}](r - r_{0})^2
+                  + U_{0} + U_{\rm WCA}(r)      & \text{when } r < r_0\\
+             = & U_0                            & \text{when } r\ge r_0
+        \end{eqnarray*}
+
+    where :math:`\vec{r}` is the vector pointing from one particle to the other in the
+    bond.
+    The potential :math:`U_{\rm WCA}(r)` is given by:
```
mphoward, line 83:
```suggestion
    bond. The potential :math:`U_{\rm WCA}(r)` is given by:
```

src/bond.py
```diff
@@ -63,3 +63,100 @@ def __init__(self):
...
+
+    `Quartic` specifies a quartic potential between the two particles in
+    the simulation state with:
+
+    .. math::
+        \begin{eqnarray*}
+        U(r) = & k [(r - r_{0}) - b_{1}][(r - r_{0}) - b_{2}](r - r_{0})^2
+                  + U_{0} + U_{\rm WCA}(r)      & \text{when } r < r_0\\
+             = & U_0                            & \text{when } r\ge r_0
+        \end{eqnarray*}
```
mphoward, line 79: I think you should include Delta in the equations explicitly. Is this right?
```suggestion
        :nowrap:

        \begin{eqnarray*}
        U(r) &= k (r - \Delta - r_0 - b_1)(r - \Delta - r_0 - b_2)(r - \Delta - r_0)^2
                  + U_0 + U_{\rm WCA}(r), & r < r_0 + \Delta \\
             &= U_0, & r \ge r_0 + \Delta
        \end{eqnarray*}
```
You can add an internal line break if the equation gets too wide.

src/bond.py
```diff
@@ -63,3 +63,100 @@ def __init__(self):
...
+
+    .. math::
+        :nowrap:
+
+        \begin{eqnarray*}
+        U_{\mathrm{WCA}}(r)= & 4 \varepsilon \left[ \left( \frac{\sigma}{r} \right)^{12}
+                             - \left( \frac{\sigma}{r} \right)^{6} \right] + \varepsilon
+                                          & \text{when } r < 2^{\frac{1}{6}}\sigma\\
+                           = & 0          & \text{when } r \ge 2^{\frac{1}{6}}\sigma
+        \end{eqnarray*}
```
mphoward, line 93:
```suggestion
        \begin{eqnarray*}
        U_{\rm WCA}(r) &= 4 \varepsilon \left[ \left( \frac{\sigma}{r-\Delta} \right)^{12}
                             - \left( \frac{\sigma}{r-\Delta} \right)^{6} \right] + \varepsilon,
                                          & r < 2^{\frac{1}{6}}\sigma + \Delta \\
                           &= 0,          & r \ge 2^{\frac{1}{6}}\sigma + \Delta
        \end{eqnarray*}
```

src/bond.py
```diff
@@ -63,3 +63,100 @@ def __init__(self):
...
+    Attributes:
+        params (TypeParameter[``bond type``, dict]):
+            The parameters of the quartic bonds for each particle type.
+            The dictionary has the following keys:
+
+            * ``k`` (`float`, **required**) - quartic attractive force strength
+              :math:`[\mathrm{energy/length^4}]`
+
+            * ``r_0`` (`float`, **required**) - Location of the quartic potential
+              cutoff. Intended to be larger than the WCA cutoff,
```
mphoward, line 105:
```suggestion
              cutoff :math:`r_0`. Intended to be larger than the WCA cutoff,
```

src/bond.py
```diff
@@ -63,3 +63,100 @@ def __init__(self):
...
+              :math:`2^{\frac{1}{6}}\sigma`. When true,
+              :math:`U(r_0) = U_{0} + U_{\rm WCA}(r_0)`
+              :math:`[\mathrm{length}]`
+
+            * ``b_1`` (`float`, **required**) - First quartic potential fitting
+              parameter
+              :math:`[\mathrm{length}]`
```
mphoward, line 112:
```suggestion
              parameter :math:`b_1` :math:`[\mathrm{length}]`.
```

src/bond.py
```diff
@@ -63,3 +63,100 @@ def __init__(self):
...
+
+            * ``b_2`` (`float`, **required**) - Second quartic potential fitting
+              parameter
+              :math:`[\mathrm{length}]`
+
```
mphoward, line 117:
```suggestion
              parameter :math:`b_2` :math:`[\mathrm{length}]`.

```

src/bond.py
```diff
@@ -63,3 +63,100 @@ def __init__(self):
...
+            * ``U_0`` (`float`, **required**) - Quartic potential energy barrier height
+              at :math:`r_0` when :math:`r_0 > 2^{\frac{1}{6}}\sigma`
+              :math:`[\mathrm{energy}]`
```
mphoward, line 120:
```suggestion
              :math:`U_0` at :math:`r_0` when :math:`r_0 > 2^{\frac{1}{6}}\sigma`
              :math:`[\mathrm{energy}]`.
```

src/bond.py
```diff
@@ -63,3 +63,100 @@ def __init__(self):
...
+
+            * ``epsilon`` (`float`, **required**) - Repulsive WCA force strength
+              parameter
+              :math:`[\mathrm{energy}]`
```
mphoward, line 124:
```suggestion
            * ``epsilon`` (`float`, **required**) - Repulsive WCA interaction energy
              :math:`\varepsilon` :math:`[\mathrm{energy}]`.
```

src/bond.py
```diff
@@ -63,3 +63,100 @@ def __init__(self):
...
+
+            * ``sigma`` (`float`, **required**) - Repulsive WCA interaction distance
+              parameter
+              :math:`[\mathrm{length}]`
```
mphoward, line 128:
```suggestion
            * ``sigma`` (`float`, **required**) - Repulsive WCA interaction size
              :math:`\sigma` :math:`[\mathrm{length}]`.
```

src/bond.py
```diff
@@ -63,3 +63,100 @@ def __init__(self):
...
+
+            * ``Delta`` (`float`, **optional**) - bond length shift ``Delta``.
+              Effectively replaces all instances of :math:`r \text{ with } r - \Delta`.
+              The default value is zero.
+              :math:`[\mathrm{length}]`
```
mphoward, line 133:
```suggestion
            * ``Delta`` (`float`, **optional**) - Shift :math:`\Delta`,
              defaults to zero :math:`[\mathrm{length}]`.
```

src/pytest/test_bond.py
```diff
@@ -68,6 +68,129 @@
...
+        hoomd.azplugins.bond.Quartic,
+        dict(
+            k=1434.3,
+            r_0=1.5,
+            b_1=-0.7589,
+            b_2=0,
+            U_0=67.2234,
+            sigma=0.0,
+            epsilon=0.0,
+            delta=0.0,
```
mphoward, line 84: Should we omit setting this parameter at least once if it has a default?
```suggestion
```

src/bond.py
```diff
@@ -63,3 +63,100 @@ def __init__(self):
...
+
+    Examples::
+
+        quartic = azplugins.bond.Quartic()
+        quartic.params['A-A'] = dict(k=1434.3, r_0=1.5, b_1=-0.7589, b_2=0.0,
```
mphoward, line 138: These are a rather random set of parameters. Do they come from somewhere that should be referenced / explained, or can they be simplified? Also, I know Antonia mentioned having only one example, but I would not use `delta=0` in the example if is the default. You can omit it, make this example use a nonzero delta, or make two examples (one with default delta, one without).

mphoward (changes requested):
Thank you! Overall, this looks really good. Please address my comments, and then this will be ready to merge after a final quick review.

mphoward:
I merged `azplugins-v1` into `main` today, so you will need to reopen / retarget this PR. Thanks!

src/bond.py
bcjeong1616: These parameters come from a 2004 paper by Tsige and Stevens, who arrived at these parameters by fitting a FENE bond potential to this quartic bond potential form [here](https://pubs.acs.org/doi/10.1021/ma034970t) I chose to use them here since this paper was one of the first to use this potential in the context of simulating breakable bonds with MD. They can, however, just as easily be replaced by "nicer" quantities. Which would you prefer?

Thank you for the note on the examples! I'll make two examples then.

src/pytest/test_bond.py
bcjeong1616: Yes, I agree thank you for the feedback! I did have a few test cases where delta was omitted, but these were incompatible with the current test_energy_and_force(), failing at line 222, where it asserts that the parameters are unchanged.

```python
# test that parameters are still correct after attach runs
assert potential.params['A-A'] == potential_test.params
```

Thus, prior to making the PR, I removed this assertion and confirmed that the tests passed with delta omitted locally.
I then reinserted this assertion and removed the test cases for consistency.

Do you have a suggestion for how to approach this?

src/bond.py
mphoward: OK! Please add some text and the DOI link to that paper to briefly explain this. Context is helpful for new folks reading documentation.

src/pytest/test_bond.py
mphoward: Oh OK, I see! For now, leave it as is. Eventually, we will add automatic testing of the documentation, so having an example there that does not set delta explicitly will cover this case.

PR #78, opened 2024-06-04
Refactor flow fields

mayukh33:

PR #83, opened 2024-07-19
Refactor/ flow profile

tzortziskoul:
Updates in flow.py to run simulations of flow fields, tested with hoomd-4.8.0.

Tests passed.


mphoward:
Thanks for your interest in working on the flow profiler! We have already made the modifications to the flow fields in PR #78:

https://github.com/mphowardlab/azplugins/pull/78

So, please revert any changes you have made to these.

The `FlowProfiler` should be configured as a `Compute`.

PR #86, opened 2024-07-25
Refactor/flow profile

tzortziskoul:
Add compute.py with class FlowProfile to calculate flow profiles throughout the simulation
Update documentation
Added new test_flow_profile pytest.
Tested with hoomd-4.8.0 and working

*Fix import issues
*Fix sorting order of imports for compute.py and test_flow_profile.py
*Fix formatting issues

### DIFF
--- a/src/compute.py
+++ b/src/compute.py
@@ -4,58 +4,63 @@
 
 """Flow profiles."""
 
-import hoomd
-from hoomd import _hoomd
-from hoomd.operation import Compute
-from hoomd.logging import log
-import numpy as np
+import numpy
 from hoomd.data.parameterdicts import ParameterDict
+from hoomd.logging import log
+from hoomd.operation import Compute
+
 
 class FlowProfile(Compute):
-    R"""Measure average profiles along a spatial dimension.
+    """Measure average profiles along a spatial dimension.
 
     The average density, velocity, and temperature profiles are computed along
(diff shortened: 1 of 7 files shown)


src/compute.py
```diff
@@ -0,0 +1,155 @@
...
+
+"""Flow profiles."""
+
+import numpy
+from hoomd.data.parameterdicts import ParameterDict
+from hoomd.logging import log
+from hoomd.operation import Compute
+
+
+class FlowProfile(Compute):
```
mphoward, line 13: I believe this class would need to inherit from `Action` if we want to be able to attach it to the simulation, e.g., as a `CustomWriter`. However, we could also force the user to manually call it in a for loop of `run()` because HOOMD 4 does not have a concept of a callback or an "analyzer". @astatt what do you think?

mphoward:
I merged `azplugins-v1` into `main` today, so you will need to reopen / retarget this PR. Thanks!

tzortziskoul:
Professor Howard,

Sorry for the delay in answering, I was travelling overseas, and I have some time difference right now. I will modify the class accordingly with that you said.

Best,
Tzortzis Koulaxizis
Graduate Research Assistant | Statt Lab |
The University of Illinois at Urbana-Champaign |
Department of Chemical and Biomolecular Engineering

________________________________

PR #89, opened 2024-08-02
Refactor/flow profile

tzortziskoul:
Flow Profile inherits from Action - instead of Compute


mphoward:
This PR will be superseded by a C++/CUDA implementation of flow profiling for performance.

PR #90, opened 2024-08-13
Expanded Yukawa Pair Potential

ashleyk6-code:
Implemented an Expanded Yukawa pair potential in azplugins and checked documentation.


astatt:
@ashleyk6-code Thank you for implementing this. It looks like you added ALL of the compiled documentation files to the PR, and it appears that you have changed 51 files because of this. Please remove these files - they are not necessary and are just cluttering the GitHub. Thank you!

mphoward:
You will want something like `git rm --cached doc/_build` followed by a commit. You may need to add the `-r` flag because this is a directory, I am not sure.

src/pair.py
```diff
@@ -459,3 +459,64 @@ def __init__(self, nlist, default_r_cut=None, mode='none'):
...
+
+    `ExpandedYukawa` is a shifted Yukawa potential. The discontinuity is shifted
+    based on a :math:`\Delta` that is the minimum interaction distance between
+    particles.
+
+    The `ExpandedYukawa` potential is:
+
+    .. math::
+
+        U(r) = \varepsilon \frac{\exp[-\kappa(r - \Delta)]}{r - \Delta}
```
clpetix, line 480: Are there any safeguards for when `r = \Delta`? Should this case be tested to insure behavior is as-intended ?

clpetix (changes requested):
Overall this looks good! I had a few small suggestions and questions. After those, this is good with me! Thanks!

src/pair.py
ashleyk6-code: There are no safeguards for that case. When r = \Delta, that means that the particles are touching and the potential and force should go to infinity. Looking specifically at the expanded mie hoomd source code, there is no specific error message for when two particles touch/overlap beyond the one detailing how particles are no longer inside of the box. In this case, we are handling the error in the same way hoomd does.

src/PairEvaluatorExpandedYukawa.h
```diff
@@ -0,0 +1,137 @@
...
+
+    DEVICE bool evalForceAndEnergy(Scalar& force_divr, Scalar& pair_eng, bool energy_shift)
+        {
+        // compute the force divided by r in force_divr
+        if (rsq < rcutsq && epsilon != Scalar(0))
+            {
+            Scalar r = fast::sqrt(rsq);
+            Scalar rinv = 1 / r;
+            Scalar delta_dist = r - delta;
+            Scalar rinv_delt = 1 / (r - delta);
```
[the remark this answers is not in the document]
ashleyk6-code: Thank you! I believe I fixed this one.

src/pair.py
```diff
@@ -459,3 +459,64 @@ def __init__(self, nlist, default_r_cut=None, mode='none'):
...
+        exyuk = pair.ExpandedYukawa(default_r_cut=4.0, nlist=nl)
+        # particle types A interacting
+        exyuk.params[('A', 'A')] = dict(epsilon=1.0, kappa=1.0, delta=2.0)
+
+    .. py:attribute:: params
+
+        The `ExpandedYukawa` potential parameters. The dictionary has the following
+        keys:
+
+        * ``epsilon`` (`float`, **required**) - Energy parameter :math:`epsilon`
```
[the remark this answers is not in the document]
ashleyk6-code: I think I fixed the missing backslashes in all three cases here as well as removing a small typo earlier in the script. Thank you!

src/pair.py
```diff
@@ -459,3 +459,64 @@ def __init__(self, nlist, default_r_cut=None, mode='none'):
...
+          :math:`[\mathrm{length}]^{-1}`
+        * ``delta`` (`float`, **required**) - Minimum interaction distance
+          :math:`\Delta` :math:`[\mathrm{length}]`
+
+        Type: :class:`~hoomd.data.typeparam.TypeParameter` [`tuple`
+        [``particle_type``, ``particle_type``], `dict`]
+
+    .. py:attribute:: mode
+
+        Energy shifting/smoothing mode: ``"none"``.
```
mphoward, line 506: This should allow all cutoff modes after you add the energy shifting to the evaluator, please copy from another class.

src/pytest/test_pair.py
```diff
@@ -238,6 +238,53 @@
     ),
 ]
 
+# ExpandedYukawa
+potential_tests += [
```
mphoward, line 242: Please add a unit test with energy shifting.

mphoward (changes requested):
Thank you for your work on this! I have some comments I would like you to address, but then this will be ready for merging.

mphoward:
@astatt can you and/or @ashleyk6-code work to address the minor comments above so that this can be included in the next release?

astatt:
@mphoward  - I'll look at this on the weekend and probably make the requested changes so we can close this out.

mphoward:
Superceded by #103

PR #95, opened 2024-12-06
Update to HOOMD 5

mphoward:
* Update minimum version to HOOMD 5
* Configure ruff and pre-commit consistently with HOOMD 5
* Fix compile errors with version 5
* Bump RTD dependencies


doc/conf.py
```diff
@@ -14,79 +14,79 @@
...
-        'hoomd.md._md',
-        'hoomd.metal._metal',
-        'hoomd.mpcd._mpcd',
-        'hoomd.minimize._minimize',
-        'hoomd.hpmc._jit',
-        'hoomd.hpmc._hpmc',
+        "hoomd._hoomd",
+        "hoomd.version_config",
+        "hoomd.md._md",
+        "hoomd.metal._metal",
```
astatt, line 44: Is there any part of azplugins that needs hoomd.metal? or hoomd.hpmc?

astatt (approved):
Looks good! The only thing I noticed is that when we test hoomd import, we do test if we can import hoomd.metal, and hoomd.hpmc, which I don't think we directly use in azplugins? So maybe we do not need to test the import of those.

doc/conf.py
mphoward: This code is actually for building documentation, not testing. Due to HOOMD's conversion of plugins to be only internal (at least in their provided template), it is no longer possible to compile the documentation for azplugins without a HOOMD installation. Locally, this isn't a problem since you can do it after the build & install steps, but RTD doesn't want long-running builds. iirc, I failed at even installing HOOMD from conda-forge on RTD (can't remember why).

My solution was to modify the RTD build process:
https://github.com/mphowardlab/azplugins/blob/529ad29340c714bec1398b99e7ef652f94eed236/.readthedocs.yaml#L12-L15
so that it does a Python-only installation of both HOOMD and azplugins:
https://github.com/mphowardlab/azplugins/blob/529ad29340c714bec1398b99e7ef652f94eed236/doc/mock_install.py#L13-L25
This means that anything imported in either HOOMD or azplugins that is compiled needs to be mocked. That's why this list is only added when the RTD environment variable is set.

So, most of this list needs to stay, but I think HOOMD 5 did remove some compiled code (maybe `_jit` and `_metal`?). If yes, we should remove those because they are unused (but also don't cause any harm).

astatt: How about we leave it as it is for now, and then revisit this as we change our azplugins documentation style to be consistent with hoomd?

mphoward: I pushed a commit to clean it up since I had the file open to look anyway :) So long as this passes, we will merge.

PR #97, opened 2024-12-12
Refactor implicit evaporator as harmonic barrier

kkritikak:
This PR reimplements the implicit evaporator as a more general harmonic barrier. The refactored code uses a template to reduce code duplication between geometries. This will allow, for example, adding a cylinder geometry in future.


src/CMakeLists.txt
```diff
@@ -6,11 +6,16 @@ set(_${COMPONENT_NAME}_sources
     ConstantFlow.cc
     GroupVelocityCompute.cc
     module.cc
+    MovingHarmonicPotential.cc
     ParabolicFlow.cc
+    PlanarMovingHarmonicBarrier.cc
```
mphoward, line 11: I'm really sorry... looking at these, I realized I made a mess of these names. Could you please rename one more time as:
* `HarmonicBarrier`
* `PlanarHarmonicBarrier`
* `SphericalHarmonicBarrier`

src/MovingHarmonicPotential.cc
```diff
@@ -3,50 +3,45 @@
...
-    m_pdata->getNumTypesChangeSignal()
-        .connect<ImplicitEvaporator, &ImplicitEvaporator::reallocateParams>(this);
     }
 
-ImplicitEvaporator::~ImplicitEvaporator()
-    {
-    m_pdata->getNumTypesChangeSignal()
-        .disconnect<ImplicitEvaporator, &ImplicitEvaporator::reallocateParams>(this);
-    }
+MovingHarmonicPotential::~MovingHarmonicPotential() { };
```
mphoward, line 30: `;` not needed after function definition.
```suggestion
MovingHarmonicPotential::~MovingHarmonicPotential() { }
```

src/MovingHarmonicPotential.cc
```diff
@@ -3,50 +3,45 @@
...
  * This method only checks for warnings about the virial.
  * Deriving classes should implement the actual force compute method.
  */
-void ImplicitEvaporator::computeForces(unsigned int timestep)
+void MovingHarmonicPotential::computeForces(unsigned int timestep)
     {
     PDataFlags flags = m_pdata->getFlags();
     if (!m_has_warned
-        && (flags[pdata_flag::pressure_tensor] || flags[pdata_flag::isotropic_virial]))
+        && flags[pdata_flag::pressure_tensor])
```
mphoward, line 42: This is correct! See here:
https://github.com/glotzerlab/hoomd-blue/blob/58b7ba503cadd90cc1ed65f40208a8d2de5b7b63/hoomd/ParticleData.h#L68-L78

src/MovingHarmonicPotential.cc
```diff
@@ -58,15 +53,17 @@ namespace detail
...
-void export_ImplicitEvaporator(pybind11::module& m)
+void export_MovingHarmonicPotential(pybind11::module& m)
     {
     namespace py = pybind11;
-    py::class_<ImplicitEvaporator, std::shared_ptr<ImplicitEvaporator>>(m,
-                                                                        "ImplicitEvaporator",
-                                                                        py::base<ForceCompute>())
+    py::class_<MovingHarmonicPotential, std::shared_ptr<MovingHarmonicPotential>>(m,
+                                                                                  "MovingHarmonicPotential",
+                                                                                  py::base<ForceCompute>())
```
mphoward, line 61: We recently discovered that this way of describing the inheritance of the class leads to very strange (incorrect) behavior in Python, due to some change in pybind11. I think it should be like this instead:
```suggestion
    py::class_<HarmonicBarrier, std::shared_ptr<HarmonicBarrier>, ForceCompute>(m,
                                                                                  "HarmonicBarrier")
```
You will want to propagate this everywhere.

src/MovingHarmonicPotential.cc
```diff
@@ -58,15 +53,17 @@ namespace detail
...
         .def(py::init<std::shared_ptr<SystemDefinition>, std::shared_ptr<Variant>>())
-        .def("setParams", &ImplicitEvaporator::setParams);
+        .def("setParams", &MovingHarmonicPotential::setParams);
```
mphoward, line 63: To work with the per-type Python interface, I think this is going to need to be modified to accept a data structure for this class that contains all the parameters. That data structure is also going to need a constructor that creates it from a Python dictionary (and I think one that can go the other way too).

Take a look at the methods here for what this class needs as an example:

https://github.com/glotzerlab/hoomd-blue/blob/58b7ba503cadd90cc1ed65f40208a8d2de5b7b63/hoomd/md/PotentialExternal.h#L185

and here for how the parameter class gets implemented:

https://github.com/glotzerlab/hoomd-blue/blob/58b7ba503cadd90cc1ed65f40208a8d2de5b7b63/hoomd/md/EvaluatorExternalMagneticField.h#L48-L76

src/MovingHarmonicPotentialGPU.cc
```diff
@@ -16,9 +19,9 @@ namespace azplugins
...
  * \param interf Position of the interface
  */
-ImplicitEvaporatorGPU::ImplicitEvaporatorGPU(std::shared_ptr<SystemDefinition> sysdef,
-                                             std::shared_ptr<Variant> interf)
-    : ImplicitEvaporator(sysdef, interf)
+MovingHarmonicPotentialGPU::MovingHarmonicPotentialGPU(std::shared_ptr<SystemDefinition> sysdef,
+                                                       std::shared_ptr<Variant> interf)
+    : MovingHarmonicPotential(sysdef, interf)
     {
     m_tuner.reset(new Autotuner(32, 1024, 32, 5, 100000, "implicit_evap", m_exec_conf));
```
mphoward, line 26: The way that autotuners get created and assigned has changed. See here for an example:

https://github.com/glotzerlab/hoomd-blue/blob/58b7ba503cadd90cc1ed65f40208a8d2de5b7b63/hoomd/md/PotentialExternalGPU.h#L49-L52

The member variable is also going to look different:

https://github.com/glotzerlab/hoomd-blue/blob/58b7ba503cadd90cc1ed65f40208a8d2de5b7b63/hoomd/md/PotentialExternalGPU.h#L39

src/MovingHarmonicPotentialGPU.h
```diff
@@ -41,7 +44,7 @@ class PYBIND11_EXPORT ImplicitEvaporatorGPU : public ImplicitEvaporator
      */
     virtual void setAutotunerParams(bool enable, unsigned int period)
```
mphoward, line 45: I think this is no longer needed.

src/PlanarMovingHarmonicBarrierGPU.cu
mphoward, line 1: The boilerplate in the kernel driver functions has changed a bit, please have a look at a file in HOOMD (e.g., in the MPCD folder) for how to update it.

src/PlanarMovingHarmonicBarrierGPU.cc
mphoward, line 1: The way autotuners are used has changed, take a look in HOOMD source code for an example and update as needed.

src/MovingHarmonicPotential.h
mphoward, line 1: Do we feel like the cutoff and g parameters are actually useful for us? I don't really use them, so I would be OK if we wanted to remove them in favor of simplicity, going down to just two parameters (spring constant and offset).

mphoward (commented):
Thank you Kritika! Here are some comments.

src/MovingHarmonicPotential.h
kkritikak: I also think it's better to remove them, I also asked Arash and he also said that it's OK  to remove them, currently I haven't removed them but I will remove them soon.

src/HarmonicBarrier.h
```diff
@@ -64,14 +67,42 @@ namespace azplugins
...
+        pybind11::dict toPython()
+            {
+            pybind11::dict d;
+            d["k"] = pybind11::cast(k);
+            d["offset"] = pybind11::cast(offset);
+            d["g"] = pybind11::cast(g);
+            d["cutoff"] = pybind11::cast(cutoff);
+            return d;
+            }
+        } __attribute__((aligned(16)));
```
mphoward, line 105: After you remove `g` and `cutoff`, this struct will be the same as a `Scalar2`. That means that it should be aligned to 8 bytes for 32-bit floats and 16 bytes for 64-bit floats. You can do this like in the LJ potential:
https://github.com/glotzerlab/hoomd-blue/blob/f0415ec516c84f203d99d96f226759034683221a/hoomd/md/EvaluatorPairLJ.h#L151-L155

src/HarmonicBarrier.h
```diff
@@ -82,19 +113,43 @@ class PYBIND11_EXPORT ImplicitEvaporator : public ForceCompute
...
+        }
+
+    void setParamsPython(std::string type_name, pybind11::dict params)
+        {
+        unsigned int type_idx = m_pdata->getTypeByName(type_name);
+        param_type h_params(params);
+        setParams(type_idx, h_params);
+        }
+
+    pybind11::dict getParams(std::string type_name)
```
mphoward, line 130: For completeness, I would like to see this as a Python-specific getter, then also include a C++ getter `getParams` that just returns the actual struct (the complementary method to `setParams`). `getParamsPython` can internally use `getParams` before converting it to a dict.
```suggestion
    pybind11::dict getParamsPython(std::string type_name)
```

src/HarmonicBarrierGPU.cc
```diff
@@ -0,0 +1,49 @@
...
+ * \param interf Position of the interface
+ */
+HarmonicBarrierGPU::HarmonicBarrierGPU(std::shared_ptr<SystemDefinition> sysdef,
+                                       std::shared_ptr<Variant> interf)
+    : HarmonicBarrier(sysdef, interf)
+    {
+    m_tuner.reset(new Autotuner<1>({AutotunerBase::makeBlockSizeRange(this->m_exec_conf)},
+                                   this->m_exec_conf,
+                                   "harmonic_barrier"));
+    this->m_autotuners.push_back(m_tuner);
```
mphoward, line 29: Since this is not a derived class template, `this->` is not necessary here
```suggestion
    m_tuner.reset(new Autotuner<1>({AutotunerBase::makeBlockSizeRange(m_exec_conf)},
                                  m_exec_conf,
                                   "harmonic_barrier"));
   m_autotuners.push_back(m_tuner);
```

src/PlanarHarmonicBarrier.cc
```diff
@@ -3,45 +3,48 @@
...
     }
 
 /*!
  * \param timestep Current timestep
  */
-void ImplicitPlaneEvaporator::computeForces(unsigned int timestep)
+void PlanarHarmonicBarrier::computeForces(uint64_t timestep)
     {
-    ImplicitEvaporator::computeForces(timestep);
+    HarmonicBarrier::computeForces(timestep);
```
mphoward, line 38: I don't think any of the parent methods actually do anything, so there is no need to invoke this here.
```suggestion
```

src/PlanarHarmonicBarrierGPU.cc
```diff
@@ -0,0 +1,87 @@
...
+    {
+    m_exec_conf->msg->notice(5) << "Destroying PlanarHarmonicBarrierGPU" << std::endl;
+    }
+
+/*!
+ * \param timestep Current timestep
+ */
+void PlanarHarmonicBarrierGPU::computeForces(uint64_t timestep)
+    {
+    HarmonicBarrierGPU::computeForces(timestep);
```
mphoward, line 40: Same here
```suggestion
```

src/PlanarHarmonicBarrierGPU.cuh
```diff
@@ -0,0 +1,37 @@
...
+
+namespace hoomd
+    {
+namespace azplugins
+    {
+namespace gpu
+    {
+
+//! Kernel driver to evaluate PlanarHarmonicBarrierGPU force
+cudaError_t compute_harmonic_force(Scalar4* d_force,
```
mphoward, line 24: This function and the corresponding kernel should be renamed something more descriptive & specific to the class:
```suggestion
cudaError_t compute_force_planar_harmonic_barrier(Scalar4* d_force,
```

src/external.py
```diff
@@ -0,0 +1,61 @@
...
+import hoomd
+
+from hoomd.azplugins import _azplugins
+from hoomd.data.parameterdicts import TypeParameterDict
+from hoomd.data.typeparam import TypeParameter
+from hoomd.data.parameterdicts import ParameterDict
+from hoomd.data.typeconverter import OnlyTypes, variant_preprocessing
+from hoomd.md import force
+
+class HarmonicBarrier(force.Force):
```
mphoward, line 15: You should make these separate classes that mirror the C++ side (`PlanarHarmonicBarrier` and `SphericalHarmonicBarrier`). They can both derive from a common `HarmonicBarrier` class though! You might follow a pattern like I did here:
https://github.com/mphowardlab/azplugins/blob/fb0ac2fe6cb894e8a437e934f8197897f3d6ad49/src/compute.py#L17

src/external.py
```diff
@@ -0,0 +1,61 @@
...
+
+    _ext_module = _azplugins
+#    _cpp_class_name = "PlanarHarmonicBarrier"
```
mphoward, line 18: These are not needed in our own plugin. Specify the module explicitly when you return the class name
```suggestion
```

mphoward (changes requested):
Looks great! Please proceed with removing those two arguments as well as addressing the mostly small comments below. Last thing we need is to port some of them unit tests that validate these objects.

kkritikak:
Thanks for the feedback! 🙌
I’ve addressed most of the comments by removing the cutoff and g parameters and also improved `external.py`.  have also added a new pytest for external.py. Let me know if there’s anything else to adjust. Could you please have a look whenever you have time!

src/PlanarHarmonicBarrier.cc
kkritikak: Thank you for pointing that out! I realized the parent methods actually throw a [warning](https://github.com/mphowardlab/azplugins/blob/v0.11.0/azplugins/ImplicitEvaporator.cc#L41) about the virial contributions, so I thought it might be better to keep it for now. That said, I’m not entirely sure about it—please let me know your thoughts, and I’m happy to remove it if needed!

mphoward: You are correct, I missed that! Please leave these calls in.

src/external.py
```diff
@@ -0,0 +1,188 @@
...
+    def _make_cpp_class_name(self):
+        cpp_class_name = self.__class__.__name__
+        if isinstance(self._simulation.device, hoomd.device.GPU):
+            cpp_class_name += "GPU"
+        return cpp_class_name
+
+
+class PlanarHarmonicBarrier(HarmonicBarrier):
+    r"""Applies a purely repulsive harmonic potential to particles
+        near an interface in planar geometry..
```
mphoward, line 77: The first line of a docstring should be a short description.
```suggestion
    r"""Planar harmonic barrier normal to *z*.
```

src/external.py
```diff
@@ -0,0 +1,188 @@
...
+
+    Args:
+        interface (`hoomd.variant`) : `z` position of the interface
```
mphoward, line 80: Backticks are used for cross referencing. The type should also be variant_like.
```suggestion
        interface (`hoomd.variant.variant_like`) : *z* position of the interface
```

src/external.py
```diff
@@ -0,0 +1,188 @@
...
+
+    `PlanarHarmonicBarrier` apply a purely harmonic potential in planar geometry
+    that pushes down on nonvolatile solutes. The potential is truncated at its minimum
+    and have following form:
```
mphoward, line 84:
```suggestion
    `PlanarHarmonicBarrier` apply a purely harmonic potential in planar geometry
    with a normal in the :math:`+z` direction:
```

src/external.py
```diff
@@ -0,0 +1,188 @@
...
+
+    .. math::
+
+        \begin{eqnarray*}
+        V(d) = & 0 & d < H \\
+               & \frac{\kappa}{2} (d-H)^2 & H < d
+        \end{eqnarray*}
```
mphoward, line 91: I don't think equation array is needed, but you do want to use cases to make a piecewise function. Also, use the real coordinates now that we have specialized documentation:
```suggestion
        U(z) = \begin{cases}
         0 & z \le H \\
         \dfrac{\kappa}{2} (z-H)^2 & z > H
         \end{cases}
```

src/external.py
```diff
@@ -0,0 +1,188 @@
...
+
+    Here, the interface is located at height *H*, and may change with time.
+    The effective interface position *H* may be modified per-particle-type
+    using a *offset* (*offset* is added to *H* to determine the effective *H*)
+    :math:`\kappa` is a spring constant setting the strength of the interface
+    (:math:`\kappa` is a proxy for the surface tension).
```
mphoward, line 97:
```suggestion
    Here, the interface is located at height *H*, which may change with time.
    `interface` specifies the nominal *H*, which may then be modified
    per-particle-type by adding an ``offset``. :math:`\kappa` is a spring
    constant setting the strength of the interface, similar to a surface tension.
```

src/external.py
```diff
@@ -0,0 +1,188 @@
...
+
+    Example::
+
+        interf = hoomd.variant.Ramp(A=50.0, B=25.0, t_start=100, t_ramp=1e6)
+        evap = hoomd.azplugins.external.PlanarHarmonicBarrier(interface = interf)
+
+        evap.params['S'] = dict(k=50.0, offset=0.0)
+        evap.params['B'] = dict(k=50.0*2*2, offset=0.0)
```
mphoward, line 105:
```suggestion
        evap = hoomd.azplugins.external.PlanarHarmonicBarrier(interface=interf)
        evap.params['S'] = dict(k=50.0, offset=0.0)
        evap.params['B'] = dict(k=50.0*2**2, offset=0.0)
```
Please also add a short description of what this example is implementing.

src/external.py
```diff
@@ -0,0 +1,188 @@
...
+        keys:
+
+        * ``k`` (`float`, **required**) - Spring constant :math:`k`
+          :math:`[\mathrm{energy} \cdot \mathrm{length}^{-2}]`
+        * ``offset`` (`float`, **required**) - Shift amount per-particle-type :math:`H`
+          :math:`[\mathrm{length}]`
+
+
+        Type: :class:`~hoomd.data.typeparam.TypeParameter` [`tuple`
+        [``particle_type``], `dict`]
```
mphoward, line 119: This should be removed since it is common to the base class.

src/external.py
```diff
@@ -0,0 +1,188 @@
...
+
+    .. warning::
+        Virial calculation has not been implemented for this model because it is
+        nonequilibrium. A warning will be raised if any calls to
+        :py:class:`hoomd.logging.Logger` are made because the logger always requests
+        the virial flags. However, this warning can be safely ignored if the
+        pressure (tensor) is not being logged or the pressure is not of interest.
```
mphoward, line 126: Move this warning to the base class.

src/external.py
```diff
@@ -0,0 +1,188 @@
...
+
+    """
+
+    def __init__(self, interface=None):
+        super().__init__(interface=interface, geometry="film")
```
mphoward, line 131: You are just going to call the parent constructor after you make the fixes above, so this can just be pass.
```suggestion
    pass
```

src/external.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+class SphericalHarmonicBarrier(HarmonicBarrier):
```
mphoward, line 134: Apply similar changes as for the planar harmonic class here.

src/pytest/test_external.py
```diff
@@ -0,0 +1,217 @@
...
+    barrier.params["A"] = dict(k=kA, offset=0.1)
+    barrier.params["B"] = dict(k=kB, offset=-0.1)
+
+    sim.operations.add(barrier)
+    #  at first step the interface will not move
+    sim.run(1)
+
+    # Test forces on particles
+    forces = barrier.forces
+    energies = barrier.energies
```
mphoward, line 82: These attributes are only valid on the root rank in MPI simulations. Please shift these inside a check on `sim.device.communicator.rank`

src/pytest/test_external.py
```diff
@@ -0,0 +1,217 @@
...
+    # particle 0 is outside interaction range
+    assert numpy.isclose(energies[0], 0.0)
+    assert numpy.isclose(forces[0][0], 0.0, atol=1e-4)
+    assert numpy.isclose(forces[0][1], 0.0, atol=1e-4)
+    assert numpy.isclose(forces[0][2], 0.0, atol=1e-4)
```
mphoward, line 87: You can use `assert_allclose` to check multiple values at once, like:
```suggestion
    assert numpy.testing.assert_allclose(forces[0], [0, 0, 0], atol=1e-4)
```

src/pytest/test_external.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+@pytest.mark.parametrize("cls", [external.SphericalHarmonicBarrier], ids=["Spherical"])
+def test_spherical_harmonic_barrier(
+    simulation_factory, cls, integrator, snap, custom_variant_fixture
+):
+    sim = simulation_factory(snap)
+    sim.operations.integrator = integrator
+
+    barrier = cls(interface=custom_variant_fixture)
```
mphoward, line 68: Since this is a test for a specific class, you don't need to parametrize the test:
```suggestion
def test_spherical_harmonic_barrier(
    simulation_factory, cls, integrator, snap, custom_variant_fixture
):
    sim = simulation_factory(snap)
    sim.operations.integrator = integrator

    barrier = hoomd.azplugins.external.SphericalHarmonicBarrier(interface=custom_variant_fixture)
```

src/pytest/test_external.py
```diff
@@ -0,0 +1,217 @@
+# Copyright (c) 2018-2020, Michael P. Howard
+# Copyright (c) 2021-2024, Auburn University
+# Part of azplugins, released under the BSD 3-Clause License.
+
+import hoomd
+from hoomd.azplugins import external
```
mphoward, line 6: I prefer not to use from imports in unittests because it lets us check that we have internally imported modules as we intend the user to be able to use them.
```suggestion
import hoomd.azplugins
```

mphoward (changes requested):
Thank you Kritika! I have some points that still need to be addressed, but overall this looks great and is close to merging. Please let me know if you have any questions.

kkritikak:
Thanks! I’ve addressed most of the comments. Please let me know if there’s anything else which can be improved.

src/external.py
```diff
@@ -0,0 +1,147 @@
...
+
+        Type: :class:`~hoomd.data.typeparam.TypeParameter` [`tuple`
+        [``particle_type``], `dict`]
+
+    .. warning::
+        Virial calculation has not been implemented for this model because it is
+        nonequilibrium. A warning will be raised if any calls to
+        :py:class:`hoomd.logging.Logger` are made because the logger always requests
+        the virial flags. However, this warning can be safely ignored if the
+        pressure (tensor) is not being logged or the pressure is not of interest.
```
mphoward, line 44:
```suggestion
        Virial calculation has not been implemented for this external field.
```

src/external.py
```diff
@@ -0,0 +1,147 @@
...
+
+    Example::
+
+        # moving interface from H = 50. to H = 25.
+        interf = hoomd.variant.Ramp(A=50.0, B=25.0, t_start=100, t_ramp=1e6)
+        evap = hoomd.azplugins.external.PlanarHarmonicBarrier(interface=interf)
+        # small particle has diameter 1.0
+        evap.params['S'] = dict(k=50.0, offset=0.0)
+        # big particle is twice as large (diameter 2.0), so coefficient is scaled
+        evap.params['B'] = dict(k=50.0*2*2, offset=0.0)
```
mphoward, line 108:
```suggestion
        evap.params['S'] = dict(k=50.0, offset=-0.5)
        # big particle is twice as large (diameter 2.0), so coefficients are scaled
        evap.params['B'] = dict(k=50.0*2*2, offset=-1.0)
```

src/external.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+class PlanarHarmonicBarrier(HarmonicBarrier):
+    r"""Planar harmonic barrier normal to *z*.
+
+    Args:
+        interface (`hoomd.variant.variant_like`) : *z* position of the interface
+
+    `PlanarHarmonicBarrier` apply a purely harmonic potential in planar geometry
+    with a normal in the :math:`+z` direction:
```
mphoward, line 86:
```suggestion
    `PlanarHarmonicBarrier` applies a purely harmonic potential in a planar
    geometry with a normal in the :math:`z` direction. Particles are pushed
    in the :math:`-z` direction when they are above the `interface`:
```

src/external.py
```diff
@@ -0,0 +1,147 @@
...
+
+class HarmonicBarrier(force.Force):
+    r"""Applies harmonic barrier.
+
+    This class should not be instantiated directly. Use a derived type.
+
+    Args:
+        interface (`hoomd.variant.variant_like`): position of the interface
+
+    .. py:attribute:: params
```
mphoward, line 24: Interface also needs to be documented as an attribute.

src/external.py
```diff
@@ -0,0 +1,147 @@
...
+from hoomd.md import force
+
+
+class HarmonicBarrier(force.Force):
+    r"""Applies harmonic barrier.
+
+    This class should not be instantiated directly. Use a derived type.
+
+    Args:
+        interface (`hoomd.variant.variant_like`): position of the interface
```
mphoward, line 22:
```suggestion
        interface (`hoomd.variant.variant_like`): Position of the interface
```

src/external.py
```diff
@@ -0,0 +1,147 @@
...
+
+from hoomd.azplugins import _azplugins
+from hoomd.data.parameterdicts import TypeParameterDict
+from hoomd.data.typeparam import TypeParameter
+from hoomd.data.parameterdicts import ParameterDict
+from hoomd.md import force
+
+
+class HarmonicBarrier(force.Force):
+    r"""Applies harmonic barrier.
```
mphoward, line 17:
```suggestion
    r"""Repulsive barrier implemented as a harmonic potential.
```

src/external.py
```diff
@@ -0,0 +1,147 @@
...
+
+    .. math::
+
+        U(z) = \begin{cases}
+         0 & z \le H \\
+         \dfrac{\kappa}{2} (z-H)^2 & z > H
+        \end{cases}
+
+    Here, the interface is located at height *H*, which may change with time.
+    ``interface`` specifies the nominal *H*, which may then be modified
```
mphoward, line 96:
```suggestion
    `interface` specifies the nominal *H*, which may then be modified
```

src/external.py
```diff
@@ -0,0 +1,147 @@
...
+    """
+
+    pass
+
+
+class SphericalHarmonicBarrier(HarmonicBarrier):
+    r"""Spherical harmonic barrier.
+
+    Args:
+        interface (`hoomd.variant.variant_like`) : Radius of spherical interface.
```
mphoward, line 119:
```suggestion
        interface (`hoomd.variant.variant_like`) : Radius of sphere.
```

src/external.py
```diff
@@ -0,0 +1,147 @@
...
+
+    `SphericalHarmonicBarrier` applies a purely harmonic potential to particles
+    near a spherical interface that pushes particles inward toward the center
+    of the sphere. The potential is truncated at its minimum
+    and has the following form:
```
mphoward, line 124:
```suggestion
    `SphericalHarmonicBarrier` applies a purely harmonic potential to particles
    outside the radius of a sphere, acting to push them inward:
```

src/external.py
```diff
@@ -0,0 +1,147 @@
...
+
+    .. math::
```
mphoward, line 126: Fix the documentation from here to the end to make it mirror the planar geometry documentation. Those changes were not applied here.

mphoward (changes requested):
Thanks Kritika! There are still some issues with  the  Python documentation. Can you please carefully go through these and make sure you have implemented everything that is needed? Especially keep an eye on the sphere class, since I think the changes for the planar geometry were not propagated there.

src/external.py
kkritikak: I wanted to inform you that a warning is being raised during documentation compilation because there is no existing cross-reference for interface. As a result, Sphinx throws the following warning, causing the ReadTheDocs tests to fail:
`hoomd/azplugins/external.py:docstring of hoomd.azplugins.external.PlanarHarmonicBarrier:6: WARNING: 'any' reference target not found: interface [ref.any]`
If I enclose the interface in double backticks (interface) it doesn't cause any problem. Is there a better way to resolve this issue?

mphoward: Oh, yes, that is because that attribute is in the base class, I think, and is not inherited in the documentation. Yes, it is OK to use the double backticks to fix that warning.

src/PlanarBarrierEvaluator.h
```diff
@@ -0,0 +1,67 @@
...
+    HOSTDEVICE Scalar4 operator()(const Scalar3& pos, Scalar k, Scalar offset) const
+        {
+        const Scalar dy = pos.y - (m_H + offset);
+        // don't compute if below minimum of potential
+        if (dy <= Scalar(0.0))
+            {
+            return make_scalar4(0, 0, 0, 0);
+            }
+
+        const Scalar f = -k * dy;               // z component of force
```
mphoward, line 45:
```suggestion
        const Scalar f = -k * dy;               // y component of force
```

doc/module-azplugins-external.rst
```diff
@@ -0,0 +1,27 @@
...
+    :nosignatures:
+
+    HarmonicBarrier
+    PlanarHarmonicBarrier
+    SphericalHarmonicBarrier
+
+.. rubric:: Details
+
+.. automodule:: hoomd.azplugins.external
+    :synopsis: External barrier.
```
mphoward, line 22:
```suggestion
    :synopsis: External potentials.
```

src/SphericalBarrierEvaluator.h
```diff
@@ -0,0 +1,71 @@
...
+    HOSTDEVICE bool valid(const BoxDim& box) const
+        {
+        const Scalar3 nearest_plane_distance = box.getNearestPlaneDistance();
+        const Scalar two_R = Scalar(2.0) * m_R;
+        return (m_R >= Scalar(0.0) && nearest_plane_distance.x >= two_R
+                && nearest_plane_distance.y >= two_R && nearest_plane_distance.z >= two_R);
+        }
+
+    private:
+    Scalar m_R; // Radius of sphere
```
mphoward, line 63:
```suggestion
    Scalar m_R; //<! Radius of sphere
```

src/PlanarBarrierEvaluator.h
```diff
@@ -0,0 +1,67 @@
...
+    //! Validate the harmonic barrier location is inside the global box
+    HOSTDEVICE bool valid(const BoxDim& box) const
+        {
+        const Scalar3 lo = box.makeCoordinates(make_scalar3(0, 0, 0));
+        const Scalar3 hi = box.makeCoordinates(make_scalar3(1, 1, 1));
+        return (m_H >= lo.y && m_H < hi.y);
+        }
+
+    private:
+    Scalar m_H; // y position of interface
```
mphoward, line 59:
```suggestion
    Scalar m_H; //<! y position of interface
```

src/external.py
```diff
@@ -0,0 +1,154 @@
+# Copyright (c) 2018-2020, Michael P. Howard
+# Copyright (c) 2021-2024, Auburn University
+# Part of azplugins, released under the BSD 3-Clause License.
+
+"""External harmonic potentials."""
```
mphoward, line 5:
```suggestion
"""External potentials."""
```

src/external.py
```diff
@@ -0,0 +1,154 @@
...
+    r"""Repulsive barrier implemented as a harmonic potential.
+
+    This class should not be instantiated directly. Use a derived type.
+
+    Args:
+        location (`hoomd.variant.variant_like`): Location of the barrier.
+
+    .. py:attribute:: location
+
+        Location of the barrier. The meaning of this position is interpreted
```
mphoward, line 26:
```suggestion
        Location of the barrier. The meaning of this location is interpreted
```

src/external.py
```diff
@@ -0,0 +1,154 @@
...
+class PlanarHarmonicBarrier(HarmonicBarrier):
+    r"""Planar harmonic barrier normal to *y*.
+
+    Args:
+        location (`hoomd.variant.variant_like`): *y* position of the
+            barrier.
+
+    `PlanarHarmonicBarrier` applies a purely harmonic potential in a planar
+    geometry with a normal in the :math:`y` direction. Particles are pushed
+    in the :math:`-y` direction when they are above the ``location``:
```
mphoward, line 92:
```suggestion
    `PlanarHarmonicBarrier` applies a purely repulsive harmonic potential
    in a planar geometry with a normal in the :math:`y` direction. Particles
    are pushed in the :math:`-y` direction when they are above the
    ``location``:
```

src/external.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+class SphericalHarmonicBarrier(HarmonicBarrier):
+    r"""Spherical harmonic barrier.
+
+    Args:
+        location (`hoomd.variant.variant_like`): Radius of sphere.
+
+    `SphericalHarmonicBarrier` applies a purely harmonic potential to particles
+    outside the radius of a sphere, acting to push them inward:
```
mphoward, line 128:
```suggestion
    `SphericalHarmonicBarrier` applies a purely repulsive harmonic potential to
    particles outside the radius of a sphere, acting to push them inward:
```

mphoward (approved):
Thank you for your work on this! I refactored your code to use a template class to reduce duplication and number of files we have to maintain. I also fixed some issues with the documentation and a deadlock when the unit tests were run with MPI. I will batch commit the small changes I marked here during review, and merge after the unit tests pass. I appreciate you taking care of this!

PR #98, opened 2024-12-31
Velocity field compute

mphoward:
This PR adds the ability to compute a 3d velocity field using up to 3d histogramming. Different coordinate systems can be supported via templating, and Cartesian and cylindrical are currently supported.

PR #99, opened 2025-01-03
Center-of-mass velocity compute

mphoward:
This PR reimplements a compute for the center-of-mass velocity of a group of particles on the CPU. Additionally, it adds support for using MPCD particles in this calculation, as well as a GPU implementation.

This is related to and based on #98 and so should be rebased and merged after that.


src/ParticleDataLoader.h
```diff
@@ -50,6 +50,29 @@ class LoadHOOMDGroupPositionVelocityMass
     const unsigned int* const m_indexes;
     };
 
+//! Load HOOMD particle in a group from an index
+class LoadHOOMDGroupVelocityMass
```
astatt, line 54: Do you want to call it "Particle" instead of "HOOMD"? Or "MDParticle"? It doesn't make a difference, it is just less immediately obvious what it is (especially if people don't look at the MPCD counterpart)... optional suggestion. feel free to ignore.

src/compute.py
```diff
@@ -15,6 +15,82 @@
 from hoomd.error import DataAccessError
 
 
+class VelocityCompute(hoomd.operation.Compute):
+    r"""Center-of-mass velocity compute.
+
+    Args:
+        filter (hoomd.filter.ParticleFilter): HOOMD particles to include.
+            The default value of `None` means no HOOMD particles are included.
+        include_mpcd_particles (bool): If `True`, include MPCD particles in
```
astatt, line 24: Do you want to include default value?

src/compute.py
```diff
@@ -15,6 +15,82 @@
...
+        self._cpp_obj = cpp_class(
+            sim.state._cpp_sys_def,
+            group,
+            self.include_mpcd_particles,
+        )
+
+        super()._attach_hook()
+
+    @hoomd.logging.log(category="sequence", requires_run=True)
+    def velocity(self):
```
astatt, line 88: Did you test that the loggable is functioning as intended?

astatt (approved):
I left some minor optional comments. I also checked that the documentation renders correctly (it does). So this looks good!

src/compute.py
mphoward: Since the default values of arguments get included in the function signatures, I've tended to not document these values in the docstrings (so I don't have to remember to update them if we change the default in future). Any defaults not in a function signature should continue to be documented explicitly, though.

PR #108, opened 2025-09-15
Compute bond without using minimum image convention

clpetix:
Compute bond without using minimum image convention to avoid issues with bonds that are longer than L/2.

### DIFF
--- a/src/PotentialBond.h
+++ b/src/PotentialBond.h
@@ -1,0 +1,391 @@
+// Copyright (c) 2018-2020, Michael P. Howard
+// Copyright (c) 2021-2025, Auburn University
+// Part of azplugins, released under the BSD 3-Clause License.
+
+// File copied from HOOMD-blue. Above needed to pass license check.
+// Copyright (c) 2009-2025 The Regents of the University of Michigan.
+// Part of HOOMD-blue, released under the BSD 3-Clause License.
+
+#include "hoomd/ForceCompute.h"
+#include "hoomd/GPUArray.h"
+#include "hoomd/MeshDefinition.h"
+#include <memory>
+
+#include <vector>
+
+/*! \file PotentialBond.h
+    \brief Declares PotentialBond
+*/
+
+#ifdef __HIPCC__
+#error This header cannot be compiled by nvcc
+#endif
+
+#include <pybind11/pybind11.h>
+
+#ifndef __POTENTIALBOND_H__
+#define __POTENTIALBOND_H__
+
+namespace hoomd
+    {
+namespace md
+    {
+/*! Bond potential with evaluator support
+
+    \ingroup computes
+*/
+template<class evaluator, class Bonds> class PotentialBond : public ForceCompute
+    {
+    public:
+    //! Param type from evaluator
+    typedef typename evaluator::param_type param_type;
+
+    //! Constructs the compute
+    PotentialBond(std::shared_ptr<SystemDefinition> sysdef);
+
+    //! Constructs the compute with external Bond data
+    PotentialBond(std::shared_ptr<SystemDefinition> sysdef,
+                  std::shared_ptr<MeshDefinition> meshdef);
+
+    //! Destructor
+    virtual ~PotentialBond();
+
+    /// Set the parameters
+    virtual void setParams(unsigned int type, const param_type& param);
+    virtual void setParamsPython(std::string type, pybind11::dict param);
+
+    /// Get the parameters
+    pybind11::dict getParams(std::string type);
+
+    /// Validate bond type
+    virtual void validateType(unsigned int type, std::string action);
+
+#ifdef ENABLE_MPI
+    //! Get ghost particle fields requested by this pair potential
+    virtual CommFlags getRequestedCommFlags(uint64_t timestep);
+#endif
+
+    protected:
+    GPUArray<param_type> m_params;      //!< Bond parameters per type
+    std::shared_ptr<Bonds> m_bond_data; //!< Bond data to use in computing bonds
+
+    //! Actually compute the forces
+    virtual void computeForces(uint64_t timestep);
+    };
+
+template<class evaluator, class Bonds>
+PotentialBond<evaluator, Bonds>::PotentialBond(std::shared_ptr<SystemDefinition> sysdef)
+    : ForceCompute(sysdef)
+    {
+    m_exec_conf->msg->notice(5) << "Constructing PotentialBond<" << evaluator::getName() << ">"
+                                << std::endl;
+    assert(m_pdata);
+
+    // access the bond data for later use
+    m_bond_data = m_sysdef->getBondData();
+
+    // allocate the parameters
+    GPUArray<param_type> params(m_bond_data->getNTypes(), m_exec_conf);
+    m_params.swap(params);
+    }
+
+template<class evaluator, class Bonds>
+PotentialBond<evaluator, Bonds>::PotentialBond(std::shared_ptr<SystemDefinition> sysdef,
+                                               std::shared_ptr<MeshDefinition> meshdef)
+    : ForceCompute(sysdef)
+    {
+    m_exec_conf->msg->notice(5) << "Constructing PotentialMeshBond<" << evaluator::getName() << ">"
+                                << std::endl;
+    assert(m_pdata);
+
+    // access the bond data for later use
+    m_bond_data = meshdef->getMeshBondData();
+
+    // allocate the parameters
+    GPUArray<param_type> params(m_bond_data->getNTypes(), m_exec_conf);
+    m_params.swap(params);
+    }
+
+template<class evaluator, class Bonds> PotentialBond<evaluator, Bonds>::~PotentialBond()
+    {
+    m_exec_conf->msg->notice(5) << "Destroying PotentialBond<" << evaluator::getName() << ">"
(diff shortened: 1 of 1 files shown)


src/PotentialBond.h
```diff
@@ -0,0 +1,391 @@
...
+        if (idx_a >= m_pdata->getN())
+            images_a = m_pdata->getImages(bond.tag[0]);
+        if (idx_b >= m_pdata->getN())
+            images_b = m_pdata->getImages(bond.tag[1]);
+
+        // unwrap particles
+        posa += images_a * lengths;
+        posb += images_b * lengths;
+
+        Scalar3 dx = posa - posb;
```
clpetix, line 274: I think this should be the right place to unwrap the positions and get the bond vector.

clpetix:
@mphoward, I've copied the file over here and made a change I think should get the unwrapped positions to compute the bond length. 

I'm fairly unfamiliar working with C++ projects. Could you explain what the process is like for getting these changes connected in and usable in HOOMD?

clpetix:
@mphoward, I've applied those changes and was able to get it to compile without error after a few small tweaks.

clpetix:
@mphoward, I have this working on 1 CPU core and added the two test scenarios we discussed. It has an issue when running under MPI. I'm not sure what is happening there. Let me know if you have any ideas on what that could be, before we start with GPU & reducing code duplication.

clpetix:
@mphoward, the parametrized tests are working fine without MPI, but something is off with the MPI versions. The MPI tests worked fine [here](https://github.com/mphowardlab/azplugins/tree/62108c9b2283f063c7dc7d16681e95c3768072fe) using the GSD package so I think it must be something with using hoomd.snapshot. 

Could you have a look and see if anything sticks out? I think I must just be overlooking something

mphoward:
Ah yes, that is expected. That property returns the forces on _all_ particles, even if they are distributed onto multiple ranks, so it is a collective call that needs to be invoked on all processors. The data is only valid on rank 0 though.

clpetix:
@mphoward, I have the GPU implementation working locally and all the tests are passing. I'm not sure what is happening with the actions here. 

This should be good for you to have another look at, when you have time!

mphoward:
> I'm not sure what is happening with the actions here.

There was some authentication error preventing the runner from pulling the Docker container it needs. I re-ran the workflow, and whatever was causing that problem seems to have resolved.

src/ImagePotentialBond.h
```diff
@@ -0,0 +1,249 @@
...
+/*! \param name Name of the class in the exported python module
+    \tparam T Evaluator type to export.
+*/
+template<class T> void export_PotentialMeshBond(pybind11::module& m, const std::string& name)
+    {
+    pybind11::class_<ImagePotentialBond<T, MeshBondData>,
+                     md::PotentialBond<T, MeshBondData>,
+                     std::shared_ptr<ImagePotentialBond<T, MeshBondData>>>(m, name.c_str())
+        .def(pybind11::init<std::shared_ptr<SystemDefinition>, std::shared_ptr<MeshDefinition>>());
+    }
```
mphoward, line 243: I don't think we have any support for meshes in azplugins, so we can just remove this export method (which is unused).
```suggestion
```

src/ImagePotentialBondGPU.cuh
```diff
@@ -0,0 +1,352 @@
...
+
+#include "hoomd/BondedGroupData.cuh"
+
+#include <assert.h>
+
+/*! \file PotentialBondGPU.cuh
+    \brief Defines templated GPU kernel code for calculating the bond forces.
+*/
+
+#ifndef __IMAGE_POTENTIAL_BOND_GPU_CUH__
```
mphoward, line 19: This needs to follow the naming convention for azplugins

src/ImagePotentialBondGPU.cuh
```diff
@@ -0,0 +1,352 @@
...
+#define __IMAGE_POTENTIAL_BOND_GPU_CUH__
+
+namespace hoomd
+    {
+namespace azplugins
+    {
+namespace kernel
+    {
```
mphoward, line 27: azplugins uses a two-tier namespace for GPU code. The data structure and kernel driver function should go in `azplugins::gpu`. The corresponding kernel should go in `azplugins::gpu::kernel`. That way, you don't have to put `gpu` or `kernel` in the function names (e.g., `compute_bond_forces` can be safely used for both). Please update.

src/ImagePotentialBondGPU.cuh
```diff
@@ -0,0 +1,352 @@
...
+        // get the bonded particle's position (MEM TRANSFER: 16 bytes)
+        Scalar4 neigh_postypej = __ldg(d_pos + cur_bond_idx);
+        Scalar3 neigh_pos = make_scalar3(neigh_postypej.x, neigh_postypej.y, neigh_postypej.z);
+
+        // get images of particles
+        const int3 img_a = d_images[idx];
+        const int3 img_b = d_images[cur_bond_idx];
+
+        // get relative vector between particles using image shift
+        const Scalar3 dx = box.shift(pos - neigh_pos, img_a - img_b);
```
mphoward, line 186: Just noting that HOOMD seems to use opposite direction for these differences on CPU and GPU (here it is `a-b`, but it is `b-a` on the CPU). That seems to be right because then the force is added below. I'm not sure what the reason is for that.

src/ImagePotentialBondGPU.h
```diff
@@ -0,0 +1,161 @@
+// Copyright (c) 2018-2020, Michael P. Howard
+// Copyright (c) 2021-2025, Auburn University
+// Part of azplugins, released under the BSD 3-Clause License.
+
+#ifndef __IMAGE_POTENTIAL_BOND_GPU_H__
```
mphoward, line 5: Update this include guard too

src/ImagePotentialBondGPU.h
```diff
@@ -0,0 +1,161 @@
+// Copyright (c) 2018-2020, Michael P. Howard
+// Copyright (c) 2021-2025, Auburn University
+// Part of azplugins, released under the BSD 3-Clause License.
+
```
mphoward, line 4: Probably all the `ImagePotentialBondGPU` files should have comments for the license along the lines of what you put in the CPU one saying we copied this from HOOMD.

src/ImagePotentialBondGPU.h
```diff
@@ -0,0 +1,161 @@
...
+*/
+template<class evaluator, class Bonds>
+class ImagePotentialBondGPU : public md::PotentialBondGPU<evaluator, Bonds>
+    {
+    public:
+    //! Inherit constructors from base class
+    using md::PotentialBondGPU<evaluator, Bonds>::PotentialBondGPU;
+
+    protected:
+    virtual void computeForces(uint64_t timestep);
```
mphoward, line 42: This needs to be overridden. You also need to reimplement the method requesting the image flags in MPI because that is not present in the class you're deriving from.
```suggestion
    void computeForces(uint64_t timestep) override;
```

src/ImagePotentialBondGPU.h
```diff
@@ -0,0 +1,161 @@
...
+    GPUArray<unsigned int> m_flags; //!< Flags set during the kernel execution
```
mphoward, line 43: Does this need to be here? I would think it's in `PotentialBondGPU`.
```suggestion
```

src/ImagePotentialBondGPU.h
```diff
@@ -0,0 +1,161 @@
...
+/*! \param name Name of the class in the exported python module
+    \tparam T Evaluator type to export.
+*/
+template<class T> void export_PotentialMeshBondGPU(pybind11::module& m, const std::string& name)
+    {
+    pybind11::class_<ImagePotentialBondGPU<T, MeshBondData>,
+                     md::PotentialBondGPU<T, MeshBondData>,
+                     std::shared_ptr<ImagePotentialBondGPU<T, MeshBondData>>>(m, name.c_str())
+        .def(pybind11::init<std::shared_ptr<SystemDefinition>, std::shared_ptr<MeshDefinition>>());
+    }
```
mphoward, line 154:
```suggestion
```

src/ImagePotentialBondGPUKernel.cu
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2018-2020, Michael P. Howard
+// Copyright (c) 2021-2025, Auburn University
+// Part of azplugins, released under the BSD 3-Clause License.
+
+// clang-format off
+#include "ImagePotentialBondGPU.cuh"
+#include "hoomd/md/EvaluatorBondHarmonic.h"
+// clang-format on
```
mphoward, line 8: Not needed when this is a real file, not one generated by CMake.
```suggestion
#include "ImagePotentialBondGPU.cuh"
#include "hoomd/md/EvaluatorBondHarmonic.h"
```

src/ImagePotentialBondGPU.h
mphoward: `protected` members are accessible to a class and its children, but not outside the class. (`public` members are accessible to anybody, while `private` members are accessible only to the class that declared them.) Without seeing your compile error, I would guess the problem is that you forgot to add `this->` in front of all instances of `m_flags`.

PR #115, opened 2025-12-02
Refactor wall potential, Lennard-Jones 9-3 potential

jinnyjc:


src/pytest/test_wall.py
```diff
@@ -0,0 +1,134 @@
...
+        0,
+    ),
+    PotentialTestCase(
+        hoomd.azplugins.wall.LJ93,
+        {"sigma": 1.0, "epsilon": 1.0, "r_cut": 3.0, "r_extrap": 0.0},
+        hoomd.wall.Sphere(radius=5, origin=(0, 0, 0), inside=True),
+        (0, 0, 2.0),
+        0,
+        0,
+    ),
```
mphoward: Since the purpose of these tests is to check the functionality of the evaluator, I think we can remove these and just test the plane. We'll assume the wall potentials work correctly for all geometries.

jinnyjc:
@mphoward I have renamed the parameter `epsilon` to `A` for the LJ93 wall potential. Please take a look when you have a moment and let me know if any further changes are needed!

mphoward (approved):
Looks great, thank you Jinny!

PR #121, opened 2026-01-26
Adding Chebyshev anisotropic pair potential

mzf0069:


src/ChebyshevTensorAnisotropicPairPotential.h
```diff
@@ -0,0 +1,130 @@
...
+#include "hoomd/Index1D.h"
+
+#include "hoomd/md/NeighborList.h"
+
+namespace hoomd
+    {
+namespace azplugins
+    {
+
+class R0Interpolator;
```
mphoward: This class will probably be called like `LinearInterpolator` or `LinearInterpolator5D`

mphoward (commented):
This is a good start! Here are some comments, let me know what questions you have.

src/ChebyshevTensorAnisotropicPairPotential.h
```diff
@@ -0,0 +1,130 @@
+// Copyright (c) 2018-2020, Michael P. Howard
+// Copyright (c) 2021-2025, Auburn University
+// Part of azplugins, released under the BSD 3-Clause License.
+
+/*!
+ * \file ChebyshevTensorAnisotropicPairPotential.h
+ * \brief Declaration of ChebyshevTensorAnisotropicPairPotential
+ */
+
+#ifndef AZPLUGINS_CHEBYSHEV_TENSOR_ANISO_PAIR_POTENTIAL_H_
```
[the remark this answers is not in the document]
mzf0069: I am more inclined to accept a general term list of polynomial degrees. This would be compatible with any sampling design we choose (for example, full tensor grids or sparse subsets).

src/ChebyshevTensorAnisotropicPairPotential.h
mzf0069: For now, I think it is best to accept the r0 data array directly and construct the interpolator internally.

src/ChebyshevTensorAnisotropicPairPotential.h
```diff
@@ -0,0 +1,130 @@
...
+    {
+
+class R0Interpolator;
+
+class PYBIND11_EXPORT ChebyshevTensorAnisotropicPairPotential : public ForceCompute
+    {
+    public:
+    //! Constructor
+    ChebyshevTensorAnisotropicPairPotential(std::shared_ptr<SystemDefinition> sysdef,
+                                            std::shared_ptr<hoomd::md::NeighborList> nlist);
```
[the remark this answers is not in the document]
mzf0069: Given the decisions above, I think the constructor should accept the approximation domain, r0_data, the term list, and the coefficients. We can infer the maximum degree per dimension and the number of terms from the term list.

src/ChebyshevTensorAnisotropicPairPotential.h
```diff
@@ -0,0 +1,123 @@
...
+#include "hoomd/md/NeighborList.h"
+
+namespace hoomd
+    {
+namespace azplugins
+    {
+
+class LinearInterpolator5D;
+
+class PYBIND11_EXPORT ChebyshevTensorAnisotropicPairPotential : public ForceCompute
```
mphoward: Based on your comments, please rename this class, the file, and the include guard to `ChebyshevAnisotropicPairPotential`
```suggestion
class PYBIND11_EXPORT ChebyshevAnisotropicPairPotential : public ForceCompute
```

src/ChebyshevTensorAnisotropicPairPotential.h
```diff
@@ -0,0 +1,123 @@
...
+    {
+    public:
+    //! Constructor
+    ChebyshevTensorAnisotropicPairPotential(std::shared_ptr<SystemDefinition> sysdef,
+                                            std::shared_ptr<hoomd::md::NeighborList> nlist,
+                                            const std::array<Scalar2, 6>& domain,
+                                            const std::vector<Scalar>& r0_data,
+                                            const std::vector<unsigned int>& terms,
+                                            const std::vector<Scalar>& coeffs);
+
```
mphoward: `r0_data` should be a raw pointer to `Scalar` so that we can copy the data in-place from a NumPy array. You will need a second raw pointer or a `std::vector` that stores the dimensionality of `r0_data` along each of the 5 dimensions.

`terms` and `coeffs` also need to be raw pointers, and you'll need one more variable that is the number of terms.

Last, `domain` should probably also be a raw pointer to a regular Scalar so you can accept a NumPy array that is (6,2) in shape. Otherwise, it can be a `std::vector<Scalar2>`, and you can copy into it.

src/ChebyshevTensorAnisotropicPairPotential.h
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+    /// 6x2 domain: stored as 6 entries of Scalar2 = (min,max)
+    const GPUArray<Scalar2>& getApproximationDomain() const
+        {
+        return m_domain;
+        }
+
+    /// r0 data (theta, phi, alpha, beta, gamma) (N x 6)
+    const GPUArray<Scalar>& getR0Data() const
```
mphoward: Do you want to be expose the `R0Data` data directly after you construct the class? Same questions go for some of the members below that return `GPUArray`. You have to decide whether you are OK with the user modifying their contents or not.

src/ChebyshevTensorAnisotropicPairPotential.h
```diff
@@ -0,0 +1,123 @@
...
+        return m_coeffs;
+        }
+
+    unsigned int getNTerms() const
+        {
+        return m_Nterms;
+        }
+
+    /// Allocate storage for term list and coefficients.
+    void resizeTerms(unsigned int Nterms);
```
mphoward: Is this something we want the user to be able to do? It seems like you wouldn't want them to do that themselves, most likely. If it's a method the class needs to trigger, though, it can be protected.

src/ChebyshevTensorAnisotropicPairPotential.h
```diff
@@ -0,0 +1,123 @@
...
+
+    protected:
+    void computeForces(uint64_t timestep) override;
+
+    private:
```
mphoward: Most of the members in this section will need to be `protected` not `private` if the GPU version of the class is going to access them.

mphoward (commented):
This is going in a good direction. Let's finalize the interface, especially the constructor, then start to implement that.

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,100 @@
...
+    public:
+    //! Constructor
+    ChebyshevAnisotropicPairPotential(std::shared_ptr<SystemDefinition> sysdef,
+                                      std::shared_ptr<hoomd::md::NeighborList> nlist,
+                                      const Scalar* domain,
+                                      const Scalar* r0_data,
+                                      const unsigned int* r0_shape,
+                                      unsigned int Nterms,
+                                      const unsigned int* terms,
+                                      const Scalar* coeffs);
```
mphoward: I like to follow the convention that the number of items comes after the pointers
```suggestion
                                      const unsigned int* terms,
                                      const Scalar* coeffs,
                                      unsigned int Nterms);
```

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,100 @@
...
+
+class PYBIND11_EXPORT ChebyshevAnisotropicPairPotential : public ForceCompute
+    {
+    public:
+    //! Constructor
+    ChebyshevAnisotropicPairPotential(std::shared_ptr<SystemDefinition> sysdef,
+                                      std::shared_ptr<hoomd::md::NeighborList> nlist,
+                                      const Scalar* domain,
+                                      const Scalar* r0_data,
+                                      const unsigned int* r0_shape,
```
mphoward: Logically, should these come before the things related to the potential approximation or after? It doesn't really matter, beyond which way it makes more sense for you to read them.

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,100 @@
...
+    const GPUArray<Scalar2>& getApproximationDomain() const
+        {
+        return m_domain;
+        }
+
+    protected:
+    void computeForces(uint64_t timestep) override;
+
+    // neighbor list object
+    std::shared_ptr<hoomd::md::NeighborList> m_nlist;
```
mphoward: Member variables should be documented like this in doxygen style
```suggestion
    std::shared_ptr<hoomd::md::NeighborList> m_nlist; //!< Neighbor list
```

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,100 @@
...
+    // approximation domain (6x2): 6 rows, each is (min,max)
+    GPUArray<Scalar2> m_domain;
+
+    // intenal r0 linear interpolator
+    std::unique_ptr<LinearInterpolator5D> m_r0_interp;
+
+    // r0_data
+    GPUArray<Scalar> m_r0_data;
+
+    std::array<unsigned int, 5> m_r0_shape;
```
mphoward: We should discuss how the linear interpolator will work. It may not be necessary to store a separate variable for it, and instead just construct it each time it is needed because it will be a lightweight wrapper around the data and the shape.

Noting: if the shape is stored in an array, it will need to be copied into the interpolator to pass it to the GPU.

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,100 @@
...
+
+    // Chebyshev term list (Nterms x 6)
+    GPUArray<unsigned int> m_terms;
+
+    // coeffs (Nterms)
+    GPUArray<Scalar> m_coeffs;
+
+    // number of terms
+    unsigned int m_Nterms = 0;
```
mphoward: This doesn't need a default value assigned to it
```suggestion
    unsigned int m_Nterms;
```

mphoward (commented):
Looks good to me, let's start on the implementation!

src/pytest/test_chebyshev.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def test_attach_and_zero_force(simulation_factory, two_particle_snapshot_factory):
+    """Construct, attach, and check force/torque output."""
+
+    # Construct the Python object
+    nlist = hoomd.md.nlist.Cell(buffer=0.4)
+
+    domain = numpy.zeros((5, 2), dtype=numpy.float64)
+    terms = numpy.zeros((2, 6), dtype=numpy.uint32)
```
mphoward: I would suggest giving the domain and terms some real values, not just zeros, as it may cause problems later once we start implementing things.

src/pytest/test_chebyshev.py
```diff
@@ -0,0 +1,71 @@
...
+    assert hasattr(pot, "_cpp_obj")
+    assert pot._cpp_obj is not None
+
+    # Post-attach checks
+    assert numpy.isclose(pot.r_cut, r_cut)
+    assert pot.n_terms == 2
+    assert pot.r0_shape == (2, 2, 2, 2, 2)
+
+    assert pot._cpp_obj.n_terms == 2
+    assert numpy.isclose(pot._cpp_obj.r_cut, r_cut)
```
mphoward: `_cpp_obj` is an implementation detail, so we should not test it in detail. The check above that it exists and is not None is OK though.
```suggestion
```

src/ChebyshevAnisotropicPairPotential.cc
```diff
@@ -0,0 +1,129 @@
...
+namespace hoomd
+    {
+namespace azplugins
+    {
+
+ChebyshevAnisotropicPairPotential::ChebyshevAnisotropicPairPotential(
+    std::shared_ptr<SystemDefinition> sysdef,
+    std::shared_ptr<hoomd::md::NeighborList> nlist,
+    const Scalar* domain,
+    const float r_cut,
```
mphoward:
```suggestion
    const Scalar r_cut,
```
Replace everywhere in binding code

src/ChebyshevAnisotropicPairPotential.cc
```diff
@@ -0,0 +1,129 @@
...
+                                                                           domain.data(),
+                                                                           r_cut,
+                                                                           terms.data(),
+                                                                           coeffs.data(),
+                                                                           Nterms,
+                                                                           r0_data.data(),
+                                                                           r0_shape.data());
+            }))
+        .def_property_readonly("r_cut", &ChebyshevAnisotropicPairPotential::getRCut)
+        .def_property_readonly("n_terms", &ChebyshevAnisotropicPairPotential::getNTerms);
```
mphoward: Python prefers `num` for number, rather than `n`
```suggestion
        .def_property_readonly("num_terms", &ChebyshevAnisotropicPairPotential::getNTerms);
```

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,106 @@
...
+#include "hoomd/Index1D.h"
+
+#include "hoomd/md/NeighborList.h"
+
+namespace hoomd
+    {
+namespace azplugins
+    {
+
+template<typename T> class LinearInterpolator5D;
```
mphoward: I think this isn't needed now, right? You would include the header in the cc file if you need to make one.
```suggestion
```

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,106 @@
...
+
+class PYBIND11_EXPORT ChebyshevAnisotropicPairPotential : public ForceCompute
+    {
+    public:
+    //! Constructor
+    ChebyshevAnisotropicPairPotential(std::shared_ptr<SystemDefinition> sysdef,
+                                      std::shared_ptr<hoomd::md::NeighborList> nlist,
+                                      const Scalar* domain,
+                                      const float r_cut,
```
mphoward: Make sure to update this in this file too
```suggestion
                                      const Scalar r_cut,
```

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,106 @@
...
+    GPUArray<Scalar> m_r0_data; //!< R0 data
+
+    GPUArray<unsigned int> m_r0_shape; //!< Number of points used along each dimension to sample r0
+    };
+
+namespace detail
+    {
+///! exports to Python
+void export_ChebyshevAnisotropicPairPotential(pybind11::module& m);
+    } // end namespace detail
```
mphoward: Please check our other files, but I think we have gotten away from declaring export functions in headers to reduce compile time during development. You should be able to delete this, and just forward declare it in the module file.

src/LinearInterpolator5D.h
```diff
@@ -0,0 +1,186 @@
...
+
+#include <cassert>
+#include <cmath>
+#include <cstdint>
+
+#include "hoomd/HOOMDMath.h"
+
+#if defined(__HIPCC__) || defined(__CUDACC__)
+#define AZPLUGINS_HOSTDEVICE __host__ __device__
+#define AZPLUGINS_FORCEINLINE __forceinline__
```
mphoward, line 16: Make sure to `undef` these at the end of the header, or they will pollute the project.

src/LinearInterpolator5D.h
```diff
@@ -0,0 +1,186 @@
...
+        return idx;
+        }
+
+    AZPLUGINS_HOSTDEVICE AZPLUGINS_FORCEINLINE unsigned int size() const
+        {
+        return n0_ * n1_ * n2_ * n3_ * n4_;
+        }
+
+    private:
+    unsigned int n0_, n1_, n2_, n3_, n4_;
```
mphoward: Two things: each member variable should go on one line and HOOMD prefers to use `m_` for member variables so you should do like `m_n0` rather than `n0_`.

Maybe it would be better to make a fixed size `unsigned int[5]` and call it `m_n`?

src/LinearInterpolator5D.h
```diff
@@ -0,0 +1,186 @@
...
+            {
+            n_[d] = n[d];
+            lo_[d] = lo[d];
+            hi_[d] = hi[d];
+
+            assert(n_[d] >= 2);
+            dx_[d] = (hi_[d] - lo_[d]) / Scalar(n_[d] - 1);
+            }
+
+        nindex_ = FiveDimensionalIndex(n_[0], n_[1], n_[2], n_[3], n_[4]);
```
mphoward: If you will store the indexer, do you need the `n_` array?

src/LinearInterpolator5D.h
```diff
@@ -0,0 +1,186 @@
...
+            if (f[dim] == Scalar(n_[dim] - 1) && x[dim] == hi_[dim])
+                {
+                --bin[dim];
+                }
+
+            dloc[dim] = f[dim] - Scalar(bin[dim]);
+
+            assert(bin[dim] >= 0);
+            assert(bin[dim] < (int)n_[dim] - 1);
+            }
```
mphoward: Can these loops be fused? I am not sure you need to store f separately. It looks like it could just be computed for each dimension inside the loop.

src/LinearInterpolator5D.h
```diff
@@ -0,0 +1,186 @@
...
+
+        const Scalar xd = dloc[0];
+        const Scalar yd = dloc[1];
+        const Scalar zd = dloc[2];
+        const Scalar wd = dloc[3];
+        const Scalar vd = dloc[4];
```
mphoward: Why are we binding these to variables rather than just using them?

src/LinearInterpolator5D.h
```diff
@@ -0,0 +1,186 @@
...
+
+        Scalar c[32];
```
mphoward: It seems like the method below should work, but I am conc

Also, please make sure to add some comments documenting what's going on. If this algorithm came from somewhere, link to it. If it came from other code, we need to make sure its license is compatible with ours or we cannot merge it.

src/LinearInterpolator5D.h
```diff
@@ -0,0 +1,186 @@
...
+        for (unsigned int mask = 0; mask < 32; ++mask)
+            {
+            const unsigned int i0 = (unsigned int)(bin[0] + ((mask >> 0) & 1u));
```
mphoward: Prefer C++ style casts
```suggestion
            const unsigned int i0 = static_cast<unsigned int>(bin[0] + ((mask >> 0) & 1u));
```

src/LinearInterpolator5D.h
```diff
@@ -0,0 +1,186 @@
...
+            const unsigned int i1 = (unsigned int)(bin[1] + ((mask >> 1) & 1u));
+            const unsigned int i2 = (unsigned int)(bin[2] + ((mask >> 2) & 1u));
+            const unsigned int i3 = (unsigned int)(bin[3] + ((mask >> 3) & 1u));
+            const unsigned int i4 = (unsigned int)(bin[4] + ((mask >> 4) & 1u));
+
+            c[mask] = Scalar(data_[nindex_(i0, i1, i2, i3, i4)]);
```
mphoward: Not necessary, it will implicitly cast since c is a Scalar
```suggestion
            c[mask] =data_[nindex_(i0, i1, i2, i3, i4)];
```

src/LinearInterpolator5D.h
```diff
@@ -0,0 +1,186 @@
...
+            }
+
+        Scalar c0[16];
```
mphoward: Down here, I am worried about the amount of memory that will be allocated when we go to the GPU, as it will create huge register pressure. Is there a way to do this using only 2 arrays (say c and c0), then toggling between them? Those arrays can be longer than you need in subsequent steps, as you can always just access the smaller indexes.

src/LinearInterpolator5D.h
```diff
@@ -0,0 +1,186 @@
...
+        for (unsigned int i = 0; i < 16; ++i)
+            {
+            c0[i] = c[2 * i] * (Scalar(1) - xd) + c[2 * i + 1] * xd;
```
mphoward: Noting that it's good here to just do `Scalar(1)`. The comment about casting applies to non-constant values.

src/module.cc
```diff
@@ -69,6 +69,7 @@ void export_ParabolicFlow(pybind11::module&);
 
 // pair
 void export_AnisoPotentialPairTwoPatchMorse(pybind11::module&);
+void export_ChebyshevAnisotropicPairPotential(pybind11::module&);
```
mphoward: Ah yes, since this declaration is here, you don't need it in the other header.

src/pair.py
```diff
@@ -4,13 +4,69 @@
...
+        nterms = int(self._terms.shape[0])
+        if self._coeffs.ndim != 1 or int(self._coeffs.shape[0]) != nterms:
+            raise ValueError("coeffs must have shape (Nterms,).")
+        if self._r0_data.ndim != 5:
+            raise ValueError("r0_data must be a 5D array.")
+
+    @property
+    def r_cut(self):
+        """Cut-off distance in approximation domain"""
+        return self._r_cut
```
mphoward: You shouldn't need a property for the ones coming from C++, but you need to use the `ParameterDict` correctly for those.
```suggestion
```

src/pair.py
```diff
@@ -4,13 +4,69 @@
...
+
+    @property
+    def n_terms(self):
+        """Number of terms."""
+        return int(self._terms.shape[0])
```
mphoward: Same here, this should come from C++ automatically.
```suggestion
```

src/pair.py
```diff
@@ -4,13 +4,69 @@
...
+
+    @property
+    def r0_shape(self):
+        """r0 table shape."""
+        return tuple(int(x) for x in self._r0_data.shape)
```
mphoward: We will want to discuss how to expose the array quantities later on. However, I note for now that we don't want to have a special shape property for `r0`. It should be a NumPy inside this class, so it should work like `r0.shape`.

src/pair.py
```diff
@@ -4,13 +4,69 @@
...
 from hoomd.data.typeparam import TypeParameter
 from hoomd.md import pair
+from hoomd.md.force import Force
 from hoomd.variant import Variant
 
 
+class ChebyshevAnisotropicPairPotential(Force):
+    """Chebyshev anisotropic pair potential."""
+
+    def __init__(self, nlist, domain, terms, coeffs, r0_data, r_cut=3.0):
```
mphoward: I'm noting that `r0_data` is not the best name at the Python level. We should probably choose a more descriptive variable name. Maybe just `r0` or it could be something like `min_distance` or `r_min`.

mphoward (commented):
This is moving in a good direction! Please see my comments and let me know what questions you have.

src/ChebyshevAnisotropicPairPotential.cc
```diff
@@ -0,0 +1,435 @@
...
+    : ForceCompute(sysdef), m_nlist(nlist), m_r_cut(r_cut), m_Nterms(Nterms)
+    {
+        {
+        GPUArray<Scalar2> domain_arr(5, m_exec_conf);
+        m_domain.swap(domain_arr);
+
+        ArrayHandle<Scalar2> h_domain(m_domain, access_location::host, access_mode::readwrite);
+        for (unsigned int d = 0; d < 5; ++d)
+            {
+            h_domain.data[d] = make_scalar2(domain[2 * d], domain[2 * d + 1]);
```
mphoward: You can use an `Index2D` if you'd like to make it clearer what's happening.

src/ChebyshevAnisotropicPairPotential.cc
```diff
@@ -0,0 +1,435 @@
...
+
+    // r0_data: flat array, length = product(r0_shape)
+    size_t n_r0 = 1;
+    for (unsigned int d = 0; d < 5; ++d)
+        {
+        n_r0 *= static_cast<size_t>(r0_shape[d]);
+        }
+
+        {
+        GPUArray<Scalar> r0_arr(n_r0, m_exec_conf);
```
mphoward: Is `GPUArray` OK with a `size_t` number (no warnings)?

src/ChebyshevAnisotropicPairPotential.cc
```diff
@@ -0,0 +1,435 @@
...
+        std::copy(r0_data, r0_data + n_r0, h_r0.data);
+        }
+    }
+
+ChebyshevAnisotropicPairPotential::~ChebyshevAnisotropicPairPotential() { }
+
+void ChebyshevAnisotropicPairPotential::computeForces(uint64_t timestep)
+    {
+    // start by updating the neighborlist
+    m_nlist->compute(timestep);
```
mphoward: We will need to attach this ForceCompute as a subscriber of the neighbor list so it can tell the neighbor list what cutoff it needs, otherwise, it will not find any neighbors. Can you take a look at `PotentialPair` to see how that's done? You'll also then need to determine what the cutoff is based on the `r0` data and `r_cut`.

src/ChebyshevAnisotropicPairPotential.cc
```diff
@@ -0,0 +1,435 @@
...
+
+    Scalar lo[5];
+    Scalar hi[5];
+    for (unsigned int d = 0; d < 5; ++d)
+        {
+        lo[d] = h_domain.data[d].x;
+        hi[d] = h_domain.data[d].y;
+        }
+
+    LinearInterpolator5D<Scalar> interp(h_r0_data.data, h_r0_shape.data, lo, hi);
```
mphoward: For convenience, you may want to write an extra constructor for `LinearInterpolator5D` that can receive the domain as a `Scalar2` array, then does this copy procedure into its internal data.

src/ChebyshevAnisotropicPairPotential.cc
```diff
@@ -0,0 +1,435 @@
...
+
+    // need to start from a zero force and torque
+    m_force.zeroFill();
+    m_torque.zeroFill();
+
+    ArrayHandle<Scalar4> h_force(m_force, access_location::host, access_mode::readwrite);
+    ArrayHandle<Scalar4> h_torque(m_torque, access_location::host, access_mode::readwrite);
```
mphoward: Is this the same pattern used in `PotentialPair`? A more efficient memory access could be to grab these in overwrite, then memset to zero. But, Josh changed alot of those calls to use this `zeroFill` method, so we should follow the same pattern as him. Just double check.

src/ChebyshevAnisotropicPairPotential.cc
```diff
@@ -0,0 +1,435 @@
...
+
+    const unsigned int N = m_pdata->getN();
+
+    for (unsigned int i = 0; i < N; ++i)
+        {
+        // particle i position and orientation
+        const Scalar3 pi = make_scalar3(h_pos.data[i].x, h_pos.data[i].y, h_pos.data[i].z);
```
mphoward: This variable name is confusing because it looks like the greek letter pi. I would call it `pos_i`.

src/ChebyshevAnisotropicPairPotential.cc
```diff
@@ -0,0 +1,435 @@
...
+            assert(j < m_pdata->getN() + m_pdata->getNGhosts());
+
+            const Scalar3 pj = make_scalar3(h_pos.data[j].x, h_pos.data[j].y, h_pos.data[j].z);
+            Scalar3 dx = pi - pj;
+            // apply periodic boundary conditions
+            dx = box.minImage(dx);
+
+            // cut-off check
+            const Scalar rsq = dot(dx, dx);
+            if (rsq > rcutsq)
```
mphoward: Be careful: this `rcutsq` is the distance relative to the surface squared, not the distance between points. You'll want to use the cutoff distance you gave the neighbor list here, once you add that subscriber function.

src/ChebyshevAnisotropicPairPotential.cc
```diff
@@ -0,0 +1,435 @@
...
+                continue;
+                }
+
+            // particle j, orientation quaternion
+            const quat<Scalar> q_j(h_orientation.data[j]);
+            // dx is in lab frame, so rotate dx by conj(q_i)
+            const vec3<Scalar> dx_lab(dx.x, dx.y, dx.z);
+            const vec3<Scalar> dx_body = rotate(q_i_conj, dx_lab);
+            // relative orientation of j with respect to i
+            const quat<Scalar> q_rel = q_i_conj * q_j;
```
mphoward: Did you confirm somewhere this is the proper way to rotate back the orientation?

src/ChebyshevAnisotropicPairPotential.cc
```diff
@@ -0,0 +1,435 @@
...
+            if (rsq > rcutsq)
+                {
+                continue;
+                }
+
+            // particle j, orientation quaternion
+            const quat<Scalar> q_j(h_orientation.data[j]);
+            // dx is in lab frame, so rotate dx by conj(q_i)
+            const vec3<Scalar> dx_lab(dx.x, dx.y, dx.z);
+            const vec3<Scalar> dx_body = rotate(q_i_conj, dx_lab);
```
mphoward: I think you can do like this because `vec3` has a constructor from Scalar3?
```suggestion
            const vec3<Scalar> dx_body = rotate(q_i_conj, vec3<Scalar>(dx));
```

src/ChebyshevAnisotropicPairPotential.cc
```diff
@@ -0,0 +1,435 @@
...
+                    cosphi = Scalar(1);
+                    }
+
+                phi = std::acos(cosphi);
+                }
+
+            // get the columns of an active rotation matrix
+            const vec3<Scalar> ex = rotate(q_rel, vec3<Scalar>(1, 0, 0));
+            const vec3<Scalar> ey = rotate(q_rel, vec3<Scalar>(0, 1, 0));
+            const vec3<Scalar> ez = rotate(q_rel, vec3<Scalar>(0, 0, 1));
```
mphoward: If you're trying to get a rotation matrix, I think `VectorMath.h` supplies a `rotmat3` from a quaternion, then you can access its elements. That is probably more efficient than applying a rotation 3 times.

src/ChebyshevAnisotropicPairPotential.cc
```diff
@@ -0,0 +1,435 @@
...
+            else if (ez.z > Scalar(1))
+                {
+                beta = Scalar(0);
+                }
+            else
+                {
+                beta = std::acos(ez.z);
+                }
+
+            if (beta > Scalar(1e-7) && beta < Scalar(M_PI - 1e-7))
```
mphoward: If you are using a "magic" tolerance, we should code that into a variable so it is clear and we can adjust its meaning.

src/ChebyshevAnisotropicPairPotential.cc
```diff
@@ -0,0 +1,435 @@
...
+            else if (beta <= Scalar(1e-7))
+                {
+                alpha = Scalar(0);
+                gamma = std::atan2(ex.y, ex.x);
+                }
+            else
+                {
+                alpha = Scalar(0);
+                gamma = std::atan2(-ex.y, ex.x);
+                }
```
mphoward: I think this is the same?
```suggestion
            else
                {
                alpha = Scalar(0);
                gamma = std::atan2((beta <= Scalar(1e-7)) ? ex.y : -ex.y, ex.x);
                }
```

src/ChebyshevAnisotropicPairPotential.cc
```diff
@@ -0,0 +1,435 @@
...
+
+            if (alpha < Scalar(0))
+                {
+                alpha += Scalar(2) * M_PI;
+                }
```
mphoward: This can be moved to the first branch, it is impossible to happen in the second one

src/ChebyshevAnisotropicPairPotential.cc
```diff
@@ -0,0 +1,435 @@
...
+            else if (theta + h > hi[0])
+                {
+                dr0_dtheta = (r0 - interp(theta - h, phi, alpha, beta, gamma)) / h;
+                }
+            else
+                {
+                dr0_dtheta = (interp(theta + h, phi, alpha, beta, gamma)
+                              - interp(theta - h, phi, alpha, beta, gamma))
+                             / (Scalar(2) * h);
+                }
```
mphoward: There is a lot of code duplication here. Is it possible for you to write a method of the interpolator that will do this math on a certain variable index?

src/ChebyshevAnisotropicPairPotential.cc
```diff
@@ -0,0 +1,435 @@
...
+                {
+                dr0_dgamma = (interp(theta, phi, alpha, beta, gamma + h)
+                              - interp(theta, phi, alpha, beta, gamma - h))
+                             / (Scalar(2) * h);
+                }
+
+            // compute J
+            }
+
+        h_force.data[i].x += fi.x;
```
mphoward: Noting here that you'll need to rotate back.

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,98 @@
...
+
+    GPUArray<unsigned int> m_terms; //!< Chebyshev term list (Nterms x 6)
+
+    GPUArray<Scalar> m_coeffs; //!< Coefficients corresponding to each term
+
+    unsigned int m_Nterms; //!< Number of terms
+
+    GPUArray<Scalar> m_r0_data; //!< R0 data
+
+    GPUArray<unsigned int> m_r0_shape; //!< Number of points used along each dimension to sample r0
```
mphoward: In headers, put the variable first and logically group them using whitespace. Methods come second.

src/LinearInterpolator5D.h
```diff
@@ -0,0 +1,202 @@
...
+#define AZPLUGINS_HOSTDEVICE
+#define AZPLUGINS_FORCEINLINE inline
+#endif
+
+namespace hoomd
+    {
+namespace azplugins
+    {
+
+class FiveDimensionalIndex
```
mphoward: For consistency with HOOMD, I think you want to name it like this:
```suggestion
class Index5D
```

src/LinearInterpolator5D.h
```diff
@@ -0,0 +1,202 @@
...
+        // Compute the cell bin and fractional coordinate in each dimension.
+        int bin[5];
+        Scalar frac[5];
+
+        for (int d = 0; d < 5; ++d)
+            {
+            const unsigned int nd = m_indexer.getN(static_cast<unsigned int>(d));
+            const Scalar f = (x[d] - m_lo[d]) / m_dx[d];
+
+            int b = static_cast<int>(std::floor(static_cast<double>(f)));
```
mphoward: Can you do this without casting f to a double? That has the potential to upcast the `Scalar` if it were not a double already. I would think `floor` is fine with float or double.
```suggestion
            int b = static_cast<int>(std::floor(f));
```

src/LinearInterpolator5D.h
```diff
@@ -0,0 +1,202 @@
...
+            const unsigned int out_len = len / 2;
+
+            for (unsigned int i = 0; i < out_len; ++i)
+                {
+                out[i] = in[2 * i] * omt + in[2 * i + 1] * t;
+                }
+            // Swap input/output
+            Scalar* tmp = in;
+            in = out;
+            out = tmp;
```
mphoward, line 196: You should be able to use `swap` to do this.

src/LinearInterpolator5D.h
```diff
@@ -0,0 +1,202 @@
...
+            assert(b >= 0);
+            assert(b < static_cast<int>(nd) - 1);
+
+            bin[d] = b;
+            frac[d] = f - Scalar(b);
+            }
+
+        // Load the 2^5=32 corners of the surrounding 5D cell.
+        Scalar c0[32];
+        Scalar c[32];
```
mphoward: Is it necessary for `c` to be size 32, or can it be 16? It would also be helpful to wait to declare this array until closer to the loop where it is used, as well as for you to choose some more descriptive variable names.

mphoward (commented):
This is moving in a very good direction!

src/ChebyshevAnisotropicPairPotential.cc
mzf0069: `zeroFill` is used in `PotentialPair`. I kept it the same.

src/ChebyshevAnisotropicPairPotential.cc
mzf0069: Yes, it has a constructor from Scalar3.

src/ChebyshevAnisotropicPairPotential.cc
mzf0069: From https://www.mathworks.com/help/fusion/ug/rotations-orientation-and-quaternions.html, I am understanding that this is the proper way of handling this.

src/pytest/test_chebyshev.py
```diff
@@ -0,0 +1,695 @@
...
+
+    r0_interp = RegularGridInterpolator(
+        (theta_grid, phi_grid, alpha_grid, beta_grid, gamma_grid),
+        r0_data,
+        method="linear",
+        bounds_error=False,
+        fill_value=numpy.nan,
+    )
+
+    def run(rho, theta, phi, alpha, beta, gamma):
```
mphoward: Rather than have all these calls to create simulations internally, is it possible to parametrize this fixture on its input coordinates and expected outputs? That may help simplify the code structure.

src/pytest/test_chebyshev.py
```diff
@@ -0,0 +1,695 @@
...
+        phi=numpy.pi / 6,
+        alpha=2 * numpy.pi / 5,
+        beta=numpy.pi / 2,
+        gamma=numpy.pi / 8,
+    )
+    check_zero_pair(sim, pot)
+
+
+# Cube symmetry
+def test_chebyshev_force_torque_energy_cube_symmetry(
```
mphoward: Similarly, could this also be wrapped into the same test parametrization if you also include the class as an input?

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,715 @@
...
+        {
+        return m_Nterms;
+        }
+
+    protected:
+    // member variables
+
+    std::shared_ptr<hoomd::md::NeighborList> m_nlist; //!< Neighbor list
+
+    GPUArray<Scalar2> m_domain; //!< Approximation domain (5x2): 5 rows, each (min, max)
```
mphoward: I don't think it makes sense for this to be an array of `Scalar2` anymore. At most, it should be an array holding the upper bounds for each coordinate. In practice, it can probably just be returned from the symmetry operation, and no member variable is needed.

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,715 @@
...
+
+    Scalar m_r_cut;       //!< Cut-off distance in approximation domain
+    Scalar m_nlist_r_cut; //!< Neighbor-list cutoff = ceil(max(r0) + r_cut)
+
+    std::shared_ptr<GPUArray<Scalar>> m_r_cut_nlist; //!< r_cut matrix shared with nlist
+    bool m_attached = true;                          //!< Whether attached to the simulation
```
mphoward, line 138: Why is this being declared here? I am not used to seeing this in HOOMD code for force computes, but maybe it was added later to code that uses neighbor lists.

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,715 @@
...
+        m_domain.swap(domain_arr);
+
+        ArrayHandle<Scalar2> h_domain(m_domain, access_location::host, access_mode::readwrite);
+        const Scalar angle_lower[5] = {
+            Scalar(0.0),    // theta
+            Scalar(1.0e-5), // phi
+            Scalar(0.0),    // alpha
+            Scalar(1.0e-5), // beta
+            Scalar(0.0)     // gamma
+        };
```
mphoward: Put this in the base `ShapeSymmetry` class since it is common to all of them? Then, you can also have this class return it if you want to expose the lower bounds to the user in addition to the upper bounds.

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,715 @@
...
+
+        for (unsigned int d = 0; d < 5; ++d)
+            {
+            h_domain.data[d] = make_scalar2(angle_lower[d], ShapeSymmetryT::domain_upper[d]);
+            }
```
mphoward: I think can be removed if you remove the domain variable, as it is just a transpose of the data from the symmetry operator.

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,715 @@
...
+        }
+
+        // terms: shape (Nterms, 6), stored flat
+        {
+        GPUArray<unsigned int> terms_arr(static_cast<size_t>(Nterms) * 6, m_exec_conf);
```
mphoward: There are alot of magic 6s and 5s in the code. Should we define a static class variable that is the number of dimensions (6), then you can use that (and that -1) to make the intents of these clearer?

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,715 @@
...
+        }
+
+        // r0_shape: length 5
+        {
+        GPUArray<unsigned int> shape_arr(5, m_exec_conf);
+        m_r0_shape.swap(shape_arr);
+
+        ArrayHandle<unsigned int> h_shape(m_r0_shape,
+                                          access_location::host,
+                                          access_mode::readwrite);
```
mphoward:
```suggestion
                                          access_mode::overwrite);
```

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,715 @@
...
+        ArrayHandle<unsigned int> h_terms(m_terms, access_location::host, access_mode::readwrite);
+        std::copy(terms, terms + static_cast<size_t>(Nterms) * 6, h_terms.data);
+        }
+
+        // coeffs: shape (Nterms,)
+        {
+        GPUArray<Scalar> coeffs_arr(Nterms, m_exec_conf);
+        m_coeffs.swap(coeffs_arr);
+
+        ArrayHandle<Scalar> h_coeffs(m_coeffs, access_location::host, access_mode::readwrite);
```
mphoward:
```suggestion
        ArrayHandle<Scalar> h_coeffs(m_coeffs, access_location::host, access_mode::overwrite);
```

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,715 @@
...
+            h_domain.data[d] = make_scalar2(angle_lower[d], ShapeSymmetryT::domain_upper[d]);
+            }
+        }
+
+        // terms: shape (Nterms, 6), stored flat
+        {
+        GPUArray<unsigned int> terms_arr(static_cast<size_t>(Nterms) * 6, m_exec_conf);
+        m_terms.swap(terms_arr);
+
+        ArrayHandle<unsigned int> h_terms(m_terms, access_location::host, access_mode::readwrite);
```
mphoward:
```suggestion
        ArrayHandle<unsigned int> h_terms(m_terms, access_location::host, access_mode::overwrite);
```

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,715 @@
...
+    for (unsigned int d = 0; d < 5; ++d)
+        {
+        n_r0 *= r0_shape[d];
+        }
+
+        {
+        GPUArray<Scalar> r0_arr(n_r0, m_exec_conf);
+        m_r0_data.swap(r0_arr);
+
+        ArrayHandle<Scalar> h_r0(m_r0_data, access_location::host, access_mode::readwrite);
```
mphoward:
```suggestion
        ArrayHandle<Scalar> h_r0(m_r0_data, access_location::host, access_mode::overwrite);
```

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,715 @@
...
+        std::copy(r0_data, r0_data + n_r0, h_r0.data);
+        }
+
+    // neighbor list subscriber
+    Scalar max_r0 = *std::max_element(r0_data, r0_data + n_r0);
+    m_nlist_r_cut = std::ceil(max_r0 + m_r_cut);
+
+    m_r_cut_nlist = std::make_shared<GPUArray<Scalar>>(1, m_exec_conf);
```
mphoward: Doesn't this need to be sized large enough to be all the types in the system? Probably you need to do that here and fill them all in.

Longer term, we should make a plan for how we deal with which particles this potential applies to.

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,715 @@
...
+    unsigned int max_deg_global = 0;
+    for (unsigned int c = 0; c < 6; ++c)
+        {
+        if (max_deg[c] > max_deg_global)
+            max_deg_global = max_deg[c];
+        }
+
+    const Index2D cheb_idx(max_deg_global + 1, 6);
+    std::vector<Scalar> cheb_T_flat(cheb_idx.getNumElements());
+    std::vector<Scalar> cheb_dT_flat(cheb_idx.getNumElements());
```
mphoward: This should be done during setup since it does not change. You can add a separate method if it is needed to keep the code cleaner.

Also, noting that the Index2D solution is fine for now, but we will probably want to use a different structure later to save memory on the GPU. These storage requirements will be much larger than necessary if you have an anisotropic grid.

src/ChebyshevAnisotropicPairPotential.h
```diff
@@ -0,0 +1,715 @@
...
+    ArrayHandle<Scalar4> h_force(m_force, access_location::host, access_mode::readwrite);
+    ArrayHandle<Scalar4> h_torque(m_torque, access_location::host, access_mode::readwrite);
+
+    const unsigned int N = m_pdata->getN();
+
+    //! Euler-angle singularity tolerance for the alpha/gamma extraction.
+    const Scalar euler_singularity_tol = Scalar(1e-7);
+
+    //! beta threshold for the Jacobian (avoids 1/sin(beta) singulrity).
+    const Scalar beta_tol = Scalar(1e-5);
```
mphoward: This is coded into the lower bounds of the domain, no? Can probably remove and use that.

Or, maybe not—should the lower bounds of the domain just be the true 0s (for sampling purposes), and then the nonzero values are used here for evaluating?

mphoward (commented):
This is really nicely done! Here are some suggestions, but they don't all necessarily need to be done now. Take a look and let me know what questions you have, then I think we should try the GPU implementation if you are happy with the CPU one.

src/ChebyshevAnisotropicPairPotential.h
mzf0069: Since we are representing orientation with ZXZ Euler angles and reducing configurations in the same coordinates, the force and torque expressions in these variables become singular at beta = 0 and beta = pi, where sin(beta) = 0. Because the interpolator is also constructed on a domain that excludes those singular endpoints, I think it is safest to keep the current nonzero lower bound in the symmetry class and use that same bound consistently during evaluation.

src/ChebyshevAnisotropicPairPotential.h
mzf0069: I think `PotentialPair` is using the same pattern.

PR #124, opened 2026-03-23
Feature/dynamic bonding

astatt:
This is the current state of the dynamic bonding code which I need some feedback and help on, @mphoward. The plan is to test hoomd 5.4.0 and cuda/12.6 first before moving it to hoomd 6.0, although it doesn't compile currently on the GPU. The CPU code works and is mostly tested. 

The two big roadblocks are currently: 

-  import the correct headers for extern/neighbor. something is going wrong there that makes me unable to compile the GPU code. I marked the headers in the respective files with "@mphoward" 
- while the CPU code compiles, there is a code duplication that is necessary due to the stricter access rules on GPUArray and GPUVector. I marked the function and the two spots it shows up in the code for you to review. 

Still todo before this can be merged: 

- [x] resolve header issues with extern neighbor package
- [x] make sure GPU code works otherwise 
- [x] port unit tests, make sure they pass 
- [x] resolve CPU code duplication (or decide to be OK with it) 
- [x] port code to hoomd 6 once it compiles/works
- [x] remove all spurious print statements and comments, general code cleanup 
- [ ] larger scale systems test


src/DynamicBondUpdaterGPU.h
```diff
@@ -0,0 +1,128 @@
...
+
+#include "DynamicBondUpdater.h"
+#include "DynamicBondUpdaterGPU.cuh"
+#include "hoomd/Autotuner.h"
+
+#include "hip/hip_runtime.h"
+#include "hoomd/md/NeighborListGPUTree.cuh"
+
+// @mphoward: There is some issue with importing the right headers from the extern/neighbor
+// class. Could you please have a look at it and check which headers to import?
```
mphoward: It has been a *very* long time since I looked at this code, but we are also dusting it off now with @jinnyjc. I think that you cannot include neighbor in `.h` or `.cuh` files, or else it will pollute into the host code where it cannot be compiled. (Kind of like what happens if you include CUB in the wrong spot.)

Instead, you should be able to include `<neighbor/neighbor.h>` in the `.cu` file where it is used, but you will need a wrapper class.

src/DynamicBondUpdaterGPU.h
```diff
@@ -0,0 +1,128 @@
...
+    private:
+
+        std::shared_ptr<Autotuner<1>> m_tuner_copy_nlist;     //!< Tuner for the primitive-copy kernel
+        std::shared_ptr<Autotuner<1>> m_tuner_filter_bonds;   //!< Tuner for existing bond filter
+
+        GPUFlags<int> m_num_nonzero_bonds_flag;            //!< GPU flag for the number of valid bonds
+        GPUFlags<unsigned int> m_max_bonds_overflow_flag;  //!< GPU flag for overflow
+
+        neighbor::LBVH m_lbvh;                            //!< LBVH for group_2
+        neighbor::LBVHTraverser  m_traverser;             //!< LBVH traverer
```
mphoward: I think these need to be the wrapper objects, see:

https://github.com/glotzerlab/hoomd-blue/blob/baf6f31b2c6e6d9901ae3d9cf6123a0a61f6ca21/hoomd/md/NeighborListGPUTree.h#L71-L72

src/DynamicBondUpdaterGPU.cuh
```diff
@@ -0,0 +1,413 @@
...
+#include "hoomd/Index1D.h"
+#include "hoomd/BoxDim.h"
+#include "hoomd/ParticleData.cuh"
+#include <iostream>
+
+
+// @mphoward: There is some issue with importing the right headers from the extern/neighbor
+// class. Could you please have a look at it and check which headers to import?
+
+// #include "hip/hip_runtime.h"
```
mphoward: Yeah, no HIP for now.

src/DynamicBondUpdaterGPU.cuh
```diff
@@ -0,0 +1,413 @@
...
+                                const  Scalar r_cut,
+                                const bool groups_identical,
+                                const unsigned int N,
+                                const unsigned int block_size);
+
+cudaError_t remove_zeros_and_sort_possible_bond_array(Scalar3 *d_all_possible_bonds,
+                                                      const unsigned int size,
+                                                      int *d_max_non_zero_bonds);
+
+//! Insert operation for a point under a mapping.
```
mphoward: This stuff probably goes to the `.cu` file if we follow the pattern of the tree neighbor list.

src/DynamicBondUpdater.cc
```diff
@@ -0,0 +1,954 @@
...
+      {
+        if (h_existing_bonds_list.data[m_existing_bonds_list_indexer(tag1,i)] == tag2)
+        return true;
+      }
+      return false;
+      }
+    }
+
+
+// @mphoward: This is the function that I originally had that doesn't work anymore due to
```
mphoward: This error is most likely coming up due to a change in when an array is acquired. See the discussion here and try to backtrack:

https://github.com/glotzerlab/hoomd-blue/pull/1938

PR #127, opened 2026-05-07
Evaporation dependent perturbed Lennard Jones potential

mayukh33:


src/export_PerturbedLennardJonesEvap.cc
mphoward, line 1: This can go at the bottom of the `PerturbedLennardJonesEvap.cc` file, it doesn't need to be separate.

src/export_PerturbedLennardJonesEvapGPU.cc
mphoward, line 1: Same here for the GPU version.

src/pair.py
```diff
@@ -426,6 +429,71 @@ def __init__(self, nlist, default_r_cut=None, default_r_on=0, mode="none"):
...
+        param_dict["energy_shift"] = bool(energy_shift)
+        self._param_dict.update(param_dict)
+
+        self._domain = numpy.asarray(domain, dtype=numpy.float64)
+        self._attraction_scale_factor_data = numpy.asarray(
+            attraction_scale_factor_data, dtype=numpy.float64
+        )
+        self._attraction_scale_factor_shape = numpy.asarray(
+            attraction_scale_factor_shape, dtype=numpy.uint32
+        )
```
mphoward, line 528: I'm not positive since the arguments of this aren't documented, but can't you get the shape of the scale factor data from the array, rather than storing a separate array?

src/PerturbedLennardJonesEvap.cc
```diff
@@ -0,0 +1,239 @@
...
+    bool energy_shift,
+    const Scalar* attraction_scale_factor_data,
+    const unsigned int* attraction_scale_factor_shape,
+    const Scalar* domain,
+    std::shared_ptr<VariantInterpolated> variant)
+    : ForceCompute(sysdef), m_nlist(nlist), epsilon_x_4(params.epsilon_x_4), m_rcut(rcut),
+      m_time_scale_factor(time_scale_factor),
+      lj1(params.epsilon_x_4 * params.sigma_6 * params.sigma_6),
+      lj2(params.epsilon_x_4 * params.sigma_6), rcutsq(rcut * rcut), rwcasq(params.rwcasq),
+      sigma_6(params.sigma_6), m_energy_shift(energy_shift), m_variant(variant)
```
mphoward: The way this class is being constructed is quite confusing because of the mix of the parameter struct. Is it meant to operate on _all_ particles in the system with the same sigma and epsilon? If yes, you would be better off just passing epsilon and sigma to the class, and packing them up into a struct later if it's really needed (it probably isn't).


Also, things like `lj1`, `lj2`, `rcutsq`, `rwcasq`, and `sigma_6` are temporary variables that should be computed before they are needed not as class members.

src/PerturbedLennardJonesEvap.cc
```diff
@@ -0,0 +1,239 @@
...
+
+    auto clamp_scaled_y = [](Scalar y, Scalar height, Scalar ylo)
+    {
+        const Scalar s = (y - ylo) / (height - ylo);
+        if (!(s > Scalar(0.0)))
+            return Scalar(0.0);
+        if (s > Scalar(1.0))
+            return Scalar(1.0);
+        return s;
+    };
```
mphoward: Does this code really need to be a lambda function? It's only used once.

src/PerturbedLennardJonesEvap.cc
```diff
@@ -0,0 +1,239 @@
...
+    ArrayHandle<Scalar> h_scale_factor(scale_factor, access_location::host, access_mode::readwrite);
+
+    for (unsigned int k = 0; k < N_tot; k++)
+        {
+        const Scalar scaled_pos_y
+            = clamp_scaled_y(h_pos.data[k].y, interface_height, box.getLo().y);
+        h_scale_factor.data[k] = interp(scaled_pos_y, scaled_t);
+        }
+
+    for (unsigned int i = 0; i < N; ++i)
```
mphoward, line 207: You could reduce code duplication in this loop significantly by constructing one `PairParameterPerturbedLennardJones` outside this loop and updating its scale factor for each pair, then making a `PairEvaluatorPerturbedLennardJones` for each pair to do the calculations.

src/PerturbedLennardJonesEvap.h
```diff
@@ -0,0 +1,143 @@
...
+
+#include "LinearInterpolator2D.h"
+#include "VariantInterpolated.h"
+
+namespace hoomd
+    {
+namespace azplugins
+    {
+
+struct PairParametersPerturbedLennardJonesEvap
```
mphoward: See comment above about constructor. I think this parameter type is probably not needed unless you want to be able to set different epsilon and sigma for different pairs of particles.

src/PerturbedLennardJonesEvap.h
```diff
@@ -0,0 +1,143 @@
...
+    ~PerturbedLennardJonesEvap()
+        {
+        if (m_r_cut_nlist)
+            m_nlist->removeRCutMatrix(m_r_cut_nlist);
+        }
+
+    Scalar scaleTime(uint64_t timestep) const
+        {
+        return Scalar(static_cast<Scalar>(timestep) / m_time_scale_factor);
+        }
```
mphoward: Where is this used and does it really need to be part of the public API of the class? It looks like it could be one line of code somewhere.

src/PerturbedLennardJonesEvap.h
```diff
@@ -0,0 +1,143 @@
...
+
+    Scalar getEpsilon() const
+        {
+        return epsilon_x_4 / Scalar(4.0);
+        }
+
+    Scalar getSigma() const
+        {
+        return std::pow(sigma_6, Scalar(1.0) / Scalar(6.0));
+        }
```
mphoward: These can become simple getters / setters on epsilon and sigma if there is only one value.

src/PerturbedLennardJonesEvapGPU.cu
mphoward, line 1: Similar comments as for CPU code apply to this file.

src/variant.py
```diff
@@ -0,0 +1,27 @@
...
+# Part of azplugins, released under the BSD 3-Clause License.
+
+"""Variant."""
+
+import numpy
+import hoomd
+from hoomd.azplugins import _azplugins
+
+
+class VariantInterpolated(_azplugins.VariantInterpolated, hoomd.variant.Variant):
```
mphoward, line 12: I'm confused by the double inheritance pattern here, and I'm not sure if what it is doing is right. Did you get this from somewhere else in HOOMD?

mayukh33: Yes, scalar.py in the variant directory of HOOMD uses the same pattern.

